PR #3, opened 2025-04-22
remove loam related code from stellar-scaffold-cli

BlaineHeffron:


crates/stellar-scaffold-cli/src/commands/init.rs
```diff
@@ -175,3 +142,34 @@ fn rename_cargo_toml_remove(project: &Path, name: &str) -> Result<(), Error> {
...
+    let status = Command::new("git")
+        .args(["clone", repo_url, dest.to_str().unwrap()])
+        .status()
+        .map_err(|e| Error::GitCloneError(format!("Failed to execute git clone: {e}")))?;
+
+    if !status.success() {
+        return Err(Error::GitCloneError("Git clone command failed".to_string()));
+    }
+    Ok(())
+}
```
willemneal, line 156: Is is fine fro now, but I think it's better to use a dep like gitoxide directly so that we don't require an external dep.

willemneal (approved):

PR #5, opened 2025-04-25
add import_contract to a new stellar-scaffold-macro crate

BlaineHeffron:


crates/stellar-scaffold-macro/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+///
+/// # Panics
+///
+/// This function may panic in the following situations:
+/// - If `stellar_build::get_target_dir()` fails to retrieve the target directory
+/// - If the input tokens cannot be parsed as a valid identifier
+/// - If the directory path cannot be canonicalized
+/// - If the canonical path cannot be converted to a string
+#[proc_macro]
+pub fn import_contract(tokens: TokenStream) -> TokenStream {
```
willemneal, line 26: I think we should rename this as `import_contract_client`. It would be nice to have two macros. One for importing the local wasm. Another would let you reference the name of the deployed contract and then look up the contract id with the registry. Then we could get a client which already knows the contract id too. But definitely have that as a separate PR. That depends on the registry this just relies on the files being in the right place.

willemneal (approved):

PR #18, opened 2025-04-30
Fix: missing libs error in workflow

pselle:
Fixes the missing library error on the Check workflow, you can use this on other workflows. The workflow still fails, but that's because it's not passing Clippy.

Check out the Github Action output on this PR vs that of the main branch to see the difference.


willemneal:
Thanks! I'll make a PR to  fix the compiler errors.

willemneal (approved):

PR #24, opened 2025-05-02
feat: run npm install and build in client directories

BlaineHeffron:


willemneal (approved):
Looks good to me! We can revisit later if others have issues with it, but given that the FE is going with NPM it seems reasonable that it would be installed!

Issue #16, opened 2025-04-30
cli: account already exist error should be ignored


BlaineHeffron:
When creating an account for development / staging to be used for deploying contracts, if the account already exists in the user's aliases it will show an error `the account named 'x' already exists` - this should be ignored and the workflow should continue.

BlaineHeffron:
This issue has been resolved by upgrading to stellar-cli 22.6

PR #27, opened 2025-05-05
fix: test git action

BlaineHeffron:
Installs required deps for git action to run and adds the `fixtures` directory for testing the cli / build crate.

### DIFF
--- a/justfile
+++ b/justfile
@@ -42,8 +42,8 @@ test-integration: build-cli-test-contracts
 
 create: build
     rm -rf .soroban
-    stellar keys generate default
-    just stellar contract deploy --wasm ./target/stellar/example_status_message.wasm --alias core
+    -stellar keys generate default --fund
+    #just stellar contract deploy --wasm ./target/stellar/example_status_message.wasm --alias core --source-account default
 
 # # Builds contracts. Deploys core subcontract and then redeploys to status message.
 


willemneal:
@BlaineHeffron could you add `#[ignore]` to the failing tests? Then we can fix them in a future PR. It would nice to get one passing CI PR first.

BlaineHeffron:
@willemneal the problem is there is currently no code in `fixtures` so none of the integration tests are running. I could do a separate PR to put that in there or we could just remove it from the git action for now.

willemneal:
>  stellar keys generate default
> ⚠️ Behavior of `generate` will change in the future, and it will no longer fund by default. If you want to fund please provide `--fund` flag. If you don't need to fund your keys in the future, ignore this warning. It can be suppressed with -q flag.

One bug here. And I have another PR #35 which should fix the issues around clippy.

justfile
```diff
@@ -28,7 +28,7 @@ build:
 
 # Setup the project to use a pinned version of the CLI
 setup:
-    -cargo binstall -y --install-path ./target/bin stellar-cli --version 22.6.0
+    -cargo install stellar-cli --version 22.6.0 --root ./target
```
willemneal, line 31:
```suggestion
    -cargo install stellar-cli --debug --version 22.6.0 --root ./target
```
Will speed up the compile step. And the debug version isn't so slow.

willemneal (approved):
Just needs to pass CI and can ignore clippy for now

BlaineHeffron:
> > stellar keys generate default
> > ⚠️ Behavior of `generate` will change in the future, and it will no longer fund by default. If you want to fund please provide `--fund` flag. If you don't need to fund your keys in the future, ignore this warning. It can be suppressed with -q flag.
> 
> One bug here. And I have another PR #35 which should fix the issues around clippy.

There is an open issue to fix this here https://github.com/AhaLabs/scaffold-stellar/issues/16

PR #28, opened 2025-05-05
fix: ensure installed cli version in justfile is same as Cargo.toml

BlaineHeffron:
Installed version in justfile should match Cargo.toml version.

### DIFF
--- a/justfile
+++ b/justfile
@@ -28,7 +28,7 @@ build:
 
 # Setup the project to use a pinned version of the CLI
 setup:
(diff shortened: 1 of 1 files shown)


willemneal (approved):
Can you confirm that this works locally?

willemneal:
NVM I see that it is working on CI, but it isn't actually pulling the binary. So we should investigate upstream with why binstall isn't working. It is a known issue for a while but perhaps we can bump it's priority.

Issue #32, opened 2025-05-05
environments.toml: replace `init` with `after_deploy`; add `constructor_args`


chadoh:
We started making `environments.toml` before Soroban had constructor functions. At that point, it felt reasonable to have a single `init` script that would do a series of "initialization" calls to the contract after it was deployed.

Now that `__constructor` exists, we need a way to specify arguments to it.

We could try to override `init` to pass both `__constructor` args and to maintain its current behavior, but that's probably confusing. Instead, let's add a new `constructor_args` parameter, then rename `init` to `after_deploy`. We can alias it to `init` (for now?), so we don't break existing apps.

Example, using the `token` contract from `stellar/soroban-examples`:

```toml
constructor_args = """
--symbol SCAST --decimal 7 --name "Scaffold Stellar Token" --admin me
"""

after_deploy = """
mint --amount 2000000 --to me"
"""
```

This is needed by:

- ahalabs/scaffold-stellar-frontend#24

chadoh:
Glad you were able to stick with regular strings! From a usability perspective it's much better than the array.

BlaineHeffron:
https://github.com/AhaLabs/scaffold-stellar/pull/33

PR #33, opened 2025-05-05
feat: pass constructor args to deploy command when building clients

BlaineHeffron:
Adds `constructor_args` and changes `init` to `after_deploy` as specified in https://github.com/AhaLabs/scaffold-stellar/issues/32


crates/stellar-scaffold-cli/src/commands/build/env_toml.rs
```diff
@@ -113,10 +113,13 @@ pub struct Contract {
     pub client: bool,
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub init: Option<String>,
+    pub after_deploy: Option<String>,
```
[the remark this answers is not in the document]
BlaineHeffron: Yeah I think since its a different tool its okay at this point IMO. Also this reminds me I need to update the docs for this change.

BlaineHeffron:
> Some of the code looks a little twisty to me, and I'd prefer we got a review from @willemneal before merging. Also, are you planning to add tests?

Yes I am working on a PR that adds the fixtures directory so that I can add tests.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -440,98 +463,154 @@ export default new Client.Client({{
...
-                    if let Some(init_script) = &settings.init {
-                        eprintln!("🚀 Running initialization script for {name:?}");
-                        self.run_init_script(&name, &contract_id, init_script)
-                            .await?;
-                    }
+    async fn process_single_contract(
+        &self,
+        name: &str,
+        workspace_root: &std::path::Path,
+        contracts: Option<&IndexMap<Box<str>, env_toml::Contract>>,
```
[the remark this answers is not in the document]
BlaineHeffron: I should rename contracts to contract_settings since its just a mapping between the name and the environments.toml settings for that contract. You raise a good point, really the check for `client = true` should happen before calling `process_single_contract` so the mapping doesn't need to be passed. 

The code reorganization in this update happened because the function got over 100 lines long and clippy was complaining.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -440,98 +463,154 @@ export default new Client.Client({{
...
+        settings: Option<&env_toml::Contract>,
+        network: &Network,
+    ) -> Result<Contract, Error> {
+        if let Some(settings) = settings {
+            if let Some(id) = &settings.id {
+                return Contract::from_string(id).map_err(|_| Error::InvalidContractID(id.clone()));
+            }
+        }
+
+        let wasm_path = workspace_root.join(format!("target/stellar/{name}.wasm"));
```
willemneal, line 522: This should come from stellar-build.

willemneal:
I never gave this code a deep review the first go around. My big suggestion to improve clarity is to use the fact that `env_toml::Contract` implements default. This way you can initially `unwrap_or_default()` and then throughout the code not deal with the `Option`.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -440,98 +463,154 @@ export default new Client.Client({{
...
+        name: &str,
+        workspace_root: &std::path::Path,
+        contracts: Option<&IndexMap<Box<str>, env_toml::Contract>>,
+        network: &Network,
+        env: &str,
+    ) -> Result<(), Error> {
+        let settings = match contracts {
+            Some(contracts) => contracts.get(name),
+            None => None,
+        };
```
willemneal, line 480: Better to use a map. However, as I said above it's better to just pass the map type since it has a default and you can use it directly. Furthermore it looks like it in this example the map is passed to this method only to use a parameter to access the value in the map. So IMO it's the caller's responsibility to provide the value. In this case it should be unwrapped or default so that there is a concrete type for settings.
```suggestion
        let settings = contracts.map(|c| c.get(name));
```

BlaineHeffron: I moved it to the caller and pass the settings. Although I ended up using this to get the settings since what you showed returned an Option<Option<&Contract>>. 
```
            let settings = contracts
                .and_then(|contracts| contracts.get(name.as_str()))
                .cloned()
                .unwrap_or_default();
```

crates/stellar-scaffold-cli/src/commands/build/clients.rs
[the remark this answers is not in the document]
BlaineHeffron: The settings retrieval and client check has been moved to caller.

BlaineHeffron:
Tests have been added after merging the fixtures PR.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
BlaineHeffron: I replaced it to use stellar_build::target_dir

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,213 @@ export default new Client.Client({{
...
+        let hash = cli::contract::upload::Cmd::parse_arg_vec(&[
+            "--wasm",
+            wasm_path
+                .to_str()
+                .expect("we do not support non-utf8 paths"),
+        ])?
+        .run_against_rpc_server(None, None)
+        .await?
+        .into_result()
+        .expect("no hash returned by 'contract install'")
```
willemneal, line 558:
```suggestion
        .expect("no hash returned by 'contract upload'")
```

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,213 @@ export default new Client.Client({{
...
         if env == "production" || env == "staging" {
-            if let Some(contracts) = contracts {
-                self.handle_production_contracts(workspace_root, contracts)
-                    .await?;
+            return self.handle_production_mode(workspace_root, contracts).await;
+        }
+
+        Self::validate_contract_names(workspace_root, contracts)?;
+
+        let reordered_names = Self::reorder_package_names(&package_names, contracts);
```
willemneal, line 433: I am not familiar with the client logic. How does the order matter for the client? Are the contracts already built at this point?

BlaineHeffron: This is for running `init` scripts in the order they are specified (since one contract might depend on another to be initialized first)

BlaineHeffron: Which are now called `after_deploy`

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,212 @@ export default new Client.Client({{
...
+        env: &str,
+    ) -> Result<(), Error> {
+        let contract_id = self
+            .get_or_deploy_contract(name, workspace_root, &settings, network)
+            .await?;
+
+        // Run init script if in development or test environment
+        if (env == "development" || env == "testing") && settings.after_deploy.is_some() {
+            eprintln!("🚀 Running initialization script for {name:?}");
+            self.run_init_script(name, &contract_id, settings.after_deploy.as_ref().unwrap())
```
chadoh, line 495:
```suggestion
        // Run after_deploy script if in development or test environment
        if (env == "development" || env == "testing") && settings.after_deploy.is_some() {
            eprintln!("🚀 Running after_deploy script for {name:?}");
            self.run_after_deploy_script(name, &contract_id, settings.after_deploy.as_ref().unwrap())
```

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,212 @@ export default new Client.Client({{
...
+        workspace_root: &std::path::Path,
+        settings: &env_toml::Contract,
+        network: &Network,
+    ) -> Result<Contract, Error> {
+        if let Some(id) = &settings.id {
+            return Contract::from_string(id).map_err(|_| Error::InvalidContractID(id.clone()));
+        }
+
+        let target_dir = workspace_root.join("target");
+        let wasm_path = stellar_build::stellar_wasm_out_file(&target_dir, name);
```
chadoh, line 517: These same two lines were added in a couple places. The general DRY guideline I follow is "two copy-pastes is ok; three is not," so it might not be time to clean it up yet, but if you know this is repeated elsewhere, then maybe @willemneal can advise on a better way to map contract names to wasm paths.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,212 @@ export default new Client.Client({{
...
+                .await?;
         }
 
-        // ensure contract names are valid
+        Ok(())
+    }
+
+    async fn handle_production_mode(
+        self,
+        workspace_root: &std::path::Path,
```
chadoh, line 454: Kind of annoying that we need to pass `workspace_root` _around_ and _around_ and _around_. Smells like maybe it should be a global or an instance attribute instead.

chadoh (changes requested):
Most of my comments above are sorta nit-picky and should be fast to address. I'd also like to see the build pass before we merge this.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -422,116 +422,212 @@ export default new Client.Client({{
...
+        contracts: Option<&IndexMap<Box<str>, env_toml::Contract>>,
+    ) -> Result<(), Error> {
+        if let Some(contracts) = contracts {
+            self.handle_production_contracts(workspace_root, contracts)
+                .await?;
+        }
```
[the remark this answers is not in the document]
BlaineHeffron: Fixed

crates/stellar-scaffold-cli/src/commands/build/clients.rs
BlaineHeffron: Add a `get_wasm_path`

crates/stellar-scaffold-cli/src/commands/build/clients.rs
BlaineHeffron: I stored it on client::Args so that it is accessible without needing to pass it around. It just needs to be set before calling `run` which is done in `mod.rs`. Since `pathBuf` doesn't implement `Copy` I had to add some `clone` statements throughout the code.

BlaineHeffron:
Currently there are two issues causing tests to fail. One is this bug - https://github.com/AhaLabs/scaffold-stellar/issues/34

The other is that it seems the migration to using `stellar-build` changed the behavior of building the wasm binaries. Before, if the underlying code wasn't changed it would not rewrite the binary data. Now it rewrites it every time. This causes the hack we used to get tests to always re-install the binary to no longer work. It adds random bytes to the wasm which then just immediately gets rewritten back to its original state during the initial build step of a `build` call. This means that the deploy / binding steps will get skipped if its the second time using the same contract in a test. 

We can either change `stellar-build` to behave in the same way `loam` did or we can find another workaround for the test environment.

.github/workflows/tests.yml
```diff
@@ -35,6 +35,10 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y libudev-dev libdbus-1-dev pkg-config
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '20'
```
willemneal, line 41: This makes me feel old...

BlaineHeffron:
Tests are all passing! Currently the redeploy script is failing, perhaps a separate PR for getting that test working. Or maybe we should rework it to have the registry contract do the redeploy? Or perhaps just add tests to the registry cli?

willemneal (approved):
I'd say just remove it. We can test it in an actual test not a script. Great work!!!

willemneal:
Yay!

PR #35, opened 2025-05-06
fix: use upstream build command

willemneal:
There is work underway to support a new rust target and in general we should not repeat the logic from upstream.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/docker.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/docker.rs
@@ -1,5 +1,5 @@
-use stellar_cli::{commands as cli, CommandParser};
 use std::error::Error;
+use stellar_cli::{commands as cli, CommandParser};
 
 pub async fn start_local_stellar() -> Result<(), Box<dyn Error>> {
     let result = cli::container::StartCmd::parse_arg_vec(&["local"])?
--- a/crates/stellar-scaffold-cli/src/commands/build/docker.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/docker.rs
@@ -1,28 +1,21 @@
-use soroban_cli::{commands as cli, CommandParser};
+use stellar_cli::{commands as cli, CommandParser};
 use std::error::Error;
 
 pub async fn start_local_stellar() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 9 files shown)


crates/stellar-scaffold-cli/src/commands/build/docker.rs
```diff
@@ -1,28 +1,21 @@
...
+        {
+            eprintln!("Container is already running, proceeding to health check...");
+        } else {
+            return Err(Box::new(e));
         }
     }
-
-    wait_for_stellar_health().await?;
-    Ok(())
+    tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
BlaineHeffron, line 17: Did you intend to proceed to health check if the container is already running after waiting 10 seconds? This might cause unnecessary waiting.

BlaineHeffron (commented):
Looks good, just one question about some slightly altered logic with starting the container when run_locally is set

crates/stellar-scaffold-cli/src/commands/build/docker.rs
```diff
@@ -1,28 +1,21 @@
...
-            }
+    if let Err(e) = result {
+        if e.to_string().contains("already in use")
+            || e.to_string().contains("port is already allocated")
+        {
+            eprintln!("Container is already running, proceeding to health check...");
+        } else {
+            return Err(Box::new(e));
         }
     }
```
willemneal, line 16:
```suggestion
    if let Err(e) = result {
        if e.to_string().contains("already in use")
            || e.to_string().contains("port is already allocated")
        {
            eprintln!("Container is already running, proceeding to health check...");
        } else {
            return Err(Box::new(e));
        }
    } else {
        tokio::time::sleep(std::time::Duration::from_secs(10)).await;
    }
```

crates/stellar-scaffold-cli/src/commands/build/docker.rs
willemneal, line 17:
```suggestion
```

willemneal: Fixed it! The logic is now back to what it was.

BlaineHeffron (approved):

BlaineHeffron (approved):

PR #37, opened 2025-05-08
feat: cil update

willemneal:


BlaineHeffron (changes requested):
The justfile should also update the installed cli version.

BlaineHeffron (approved):

Issue #7, opened 2025-04-29
Registry: Add metadata to contracts during build / publish


BlaineHeffron:
Metadata needs to be added during the build and publish actions

https://github.com/orgs/stellar/discussions/1573

key: `release_name` 
value: `v<repo version string>` 

key: `source_repo`
value: `<url to repo>`

We would also need to write a github action that will call the registry-cli to build / publish the wasm based on the github version number.

pselle:
Closed by #46

PR #46, opened 2025-05-12
feat: Add metadata to wasm and use in registry publish

pselle:
Add metadata to the wasm at build time.

Add a publish command to stellar-registry-cli

https://github.com/orgs/AhaLabs/projects/5/views/1?pane=issue&itemId=106395853

addresses part of #7 by adding source_repo to the WASM, but release_name looks like a better fit for being related to Github Actions/those variables vs. in the WASM

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/mod.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/mod.rs
@@ -121,10 +121,12 @@ impl Command {
                 cmd.meta.push(("wasm_name".to_string(), p.name.clone()));
             }
             if !p.version.to_string().is_empty() {
-                cmd.meta.push(("version".to_string(), p.version.to_string()));
+                cmd.meta
+                    .push(("version".to_string(), p.version.to_string()));
             }
             if p.repository.is_some() {
-                cmd.meta.push(("source_repo".to_string(), p.repository.clone().unwrap()));
+                cmd.meta
+                    .push(("source_repo".to_string(), p.repository.clone().unwrap()));
             }
             cmd.run(&global_args)?;
         }


crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        // Read the Wasm file from the path
+        let wasm_bytes =
+            std::fs::read(&self.wasm).map_err(|_| Error::MissingFileArg(self.wasm.clone()))?;
+        let spec =
+            contract_spec::Spec::new(&wasm_bytes).map_err(|_| Error::CannotParseContractSpec);
+
+        // Get out its metadata and set the contract name (wasm_name), version, source_repo
```
pselle, line 50: @willemneal I would appreciate some Rust feedback on a smarter way to do this … This is very brute and naive (mutables, then populating them, no checks or failures when the entries aren't there…)

pselle: @willemneal I took a stab at this dea

crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }));
+
+        // Add the author argument
+        let key = self.config.key_pair()?;
+        let author = stellar_strkey::ed25519::PublicKey(key.verifying_key().to_bytes()).to_string();
+        args.push(format!("--author={author}"));
+
+        // Invoke the registry with the arguments
+        invoke_registry(&args.iter().map(std::string::String::as_str).collect::<Vec<&str>>()).await?;
```
willemneal: `String` is already in scope and you can elide the generic since the type is is the function argument.
```suggestion
        invoke_registry(&args.iter().map(String::as_str).collect::<Vec<_>>()).await?;
```

crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -0,0 +1,79 @@
...
+                    Some(format!("--{key_str}={val}"))
+                } else {
+                    None
+                }
+            }
+        }));
+
+        // Add the author argument
+        let key = self.config.key_pair()?;
+        let author = stellar_strkey::ed25519::PublicKey(key.verifying_key().to_bytes()).to_string();
```
willemneal, line 71: I would add an optional author CLI flag. It could be the cause that the signer could be a different account since you can add multiple signers to an account.

willemneal (changes requested):
Two small suggestions! Otherwise great rewrite!

willemneal (approved):

crates/stellar-scaffold-cli/src/commands/build/mod.rs
```diff
@@ -117,6 +117,17 @@ impl Command {
...
+                cmd.meta.push(("wasm_name".to_string(), p.name.clone()));
+            }
+            if !p.version.to_string().is_empty() {
+                cmd.meta
+                    .push(("version".to_string(), p.version.to_string()));
+            }
+            if p.repository.is_some() {
+                cmd.meta
+                    .push(("source_repo".to_string(), p.repository.clone().unwrap()));
+            }
```
willemneal, line 130:
```suggestion
            if let Some(repo) =  p.repository.clone() {
                cmd.meta.push(("source_repo".to_string(), p));
            }
```

PR #52, opened 2025-05-13
feat: deterministically derive contract id

willemneal:


crates/stellar-registry-cli/src/contract.rs
```diff
@@ -0,0 +1,27 @@
...
+pub fn contract_id(network_passphrase: &str) -> stellar_strkey::Contract {
+    let network_id = xdr::Hash(Sha256::digest(network_passphrase.as_bytes()).into());
+    let preimage = xdr::HashIdPreimage::ContractId(xdr::HashIdPreimageContractId {
+        network_id,
+        contract_id_preimage: xdr::ContractIdPreimage::Address(
+            xdr::ContractIdPreimageFromAddress {
+                address: xdr::ScAddress::Account(xdr::AccountId(
+                    xdr::PublicKey::PublicKeyTypeEd25519(stellar_address().0.into()),
+                )),
+                salt: xdr::Uint256([0; 32]),
```
BlaineHeffron, line 19: Did you mean the salt to be 0? I thought it was going to be "registry" or something like that?

BlaineHeffron (commented):
A comment about the salt

crates/stellar-registry-cli/src/contract.rs
willemneal: Used 0 to make it easy to test. We can change it later if we want but honestly once I went with zero it seemed even easier. No need to convert a string to bytes.

PR #64, opened 2025-05-19
feat: add stellar-registry crate

willemneal:


Cargo.toml
```diff
@@ -8,6 +8,8 @@ repository = "https://github.com/ahalabs/scaffold-stellar"
 
 [workspace.dependencies]
 stellar-scaffold-macro = { path = "crates/stellar-scaffold-macro" }
```
pselle, line 10: Is this dependency needed, when you change the macro used to stellar-registry?

BlaineHeffron (approved):

pselle:
I don't see a purpose/value-add of this crate compared to how it was, but I have no technical objection to this change.

PR #67, opened 2025-05-20
Feat: registry install

BlaineHeffron:
Fixes `stellar registry install` command. 

Also fixes `publish` and `deploy` commands. 

Additionally, `publish`, `deploy`, and `install` now all respect the flattened `config::Args` from upstream `stellar-cli`. 

Provides a basic testing framework for `stellar-registry-cli`. The test in `install` installs `hello_world` locally, install the registry contract locally, and uses the `registry-cli` to publish and deploy `hello_world`.  It then runs the `registry install` command and ensures the output is valid.


crates/stellar-registry-cli/src/commands/deploy.rs
```diff
@@ -162,12 +167,11 @@ impl Cmd {
             function_name: "deploy".try_into().unwrap(),
             args: [
                 ScVal::String(ScString(self.wasm_name.clone().try_into().unwrap())),
-                ScVal::Void,
+                ScVal::Void, // TODO should be a way to pass version
```
willemneal, line 170: Convert version by hand to ScVal

crates/stellar-registry-cli/src/commands/install.rs
```diff
@@ -33,70 +35,135 @@ pub enum Error {
...
-            .into_iter()
-            .map(Into::into)
-            .collect::<Vec<_>>();
-        println!("Fetching contract ID...\n{cmd:#?}");
-        Ok(cmd
-            .invoke(&global::Args::default())
-            .await?
-            .into_result()
-            .unwrap())
+    pub async fn get_contract_id(&self) -> Result<String, Error> {
```
willemneal, line 80: Perhaps return `stellar_strkey::Contract` here. Also you can use `Ok(contract.parse()?)`

crates/stellar-registry-cli/src/commands/install.rs
```diff
@@ -33,70 +35,135 @@ pub enum Error {
...
+            network: self.config.network.clone(),
             ..Default::default()
         };
         fetch_cmd.run().await?;
-        std::fs::write(id_file, testnet::contract_id_strkey().to_string())?;
+        println!("WASM file saved to: {}", out_file.display());
+
+        println!("✅ Successfully installed contract {}", self.contract_name);
+        println!("Contract ID: {contract_id}");
+
```
willemneal, line 76:
```suggestion
        eprintln!("WASM file saved to: {}", out_file.display());

        eprintln!("✅ Successfully installed contract {}", self.contract_name);
        eprintln!("Contract ID: {contract_id}");

```

crates/stellar-registry-cli/src/commands/install.rs
```diff
@@ -33,70 +35,135 @@ pub enum Error {
...
-            .invoke(&global::Args::default())
-            .await?
-            .into_result()
-            .unwrap())
+    pub async fn get_contract_id(&self) -> Result<String, Error> {
+        // Prepare the arguments for invoke_registry
+        let slop = vec!["fetch_contract_id", "--contract-name", &self.contract_name];
+
+        // Use this.config directly
+        println!("Fetching contract ID via registry...");
```
willemneal, line 85:
```suggestion
        eprintln!("Fetching contract ID via registry...");
```

crates/stellar-registry-cli/src/commands/install.rs
```diff
@@ -33,70 +35,135 @@ pub enum Error {
...
 impl Cmd {
     pub async fn run(&self) -> Result<(), Error> {
-        let contract_id = testnet::contract_id();
-        let network = testnet::network();
-        let out_dir = self.out_dir.as_ref().unwrap();
-        let mut out_file = out_dir.join(&self.deployed_name);
+        // Use the network config from flattened args
+        let network_passphrase = self
+            .config
+            .network
```
willemneal, line 47: You should use `self.config.get_network()` which will resolve the network including the passphrase so it's not required that it was passed explicitly.

crates/stellar-scaffold-test/src/registry.rs
```diff
@@ -0,0 +1,123 @@
...
+            "--rpc-url".to_string(),
+            "http://localhost:8000/soroban/rpc".to_string(),
+            "--network-passphrase".to_string(),
+            "Standalone Network ; February 2017".to_string(),
+            "--".to_string(),
+            "--admin".to_string(),
+            "alice".to_string(),
+        ];
+        let deploy_arg_refs: Vec<&str> = deploy_args.iter().map(String::as_str).collect();
+        let contract_id = cli::contract::deploy::wasm::Cmd::parse_arg_vec(&deploy_arg_refs)
```
willemneal, line 93:
```suggestion
        let deploy_args = [
            "--wasm-hash",
            &hash,
            "--source",
            "alice",
            "--config-dir",
            env.cwd.to_str().unwrap(),
            "--rpc-url",
            "http://localhost:8000/soroban/rpc",
            "--network-passphrase",
            "Standalone Network ; February 2017",
            "--",
            "--admin",
            "alice",
        ];
        let contract_id = cli::contract::deploy::wasm::Cmd::parse_arg_vec(&deploy_args)
```

willemneal:
Looking over it again I think we can drop fetching the wasm. I added a cache to the stellar CLI. So anytime for example you used invoke `stellar contract invoke --id alias_to_contract -- --help` It would first look up the contract id of the alias. Then look up the wasm hash of the contact ID. Then first look it up in the system wide cache, otherwise it would download the wasm and add it to the cache. I think we should reuse this same code. This way the contract will always use the correct hash and the CLI can also reuse the wasm.

willemneal:
Looked into it. It was designed with the spec in mind since that is what invoke, etc needed. However, there is `spec_dir` which is misnamed because it actually is the whole wasms. 

https://github.com/stellar/stellar-cli/blob/2ef1154aa42b3e0428cb0f821baab27f06f5a0f1/cmd/soroban-cli/src/config/data.rs#L47-L51

So perhaps we can use this directory with the location to put the wasms. And the file name is the hash. 

And we can get rid of the option of where to place the file. And for testing you can overwrite the location with an ENV see https://github.com/stellar/stellar-cli/blob/2ef1154aa42b3e0428cb0f821baab27f06f5a0f1/cmd/soroban-cli/src/config/data.rs#L28

crates/stellar-registry-cli/src/commands/deploy.rs
willemneal, line 170:
```suggestion
                self.version.clone().map(|s|ScVal::String(ScString(s.try_into().unwrap()))).unwrap_or(ScVal::Void),
```

willemneal (approved):

PR #69, opened 2025-05-21
feat(registry): validate names used in registry; to prevent potential issues when used with rust imports

willemneal:
Currently requires:
- 0 < len < 65
- ascii alphanumeric with `_`
- cannot start with number or `_`
- cannot be a rust keyword

### DIFF
--- a/contracts/registry/src/error.rs
+++ b/contracts/registry/src/error.rs
@@ -27,6 +27,6 @@ pub enum Error {
     /// New version must be greater than the most recent version
     VersionMustBeGreaterThanCurrent = 11,
     /// Invalid name.
-    /// Must be 64 characters or less; ascii alphanumeric or '_';  start with a letter; and not be a Rust keyword
+    /// Must be 64 characters or less; ascii alphanumeric or '_'; start with a letter; and not be a Rust keyword
     InvalidName = 12,
 }
--- a/contracts/registry/src/error.rs
+++ b/contracts/registry/src/error.rs
@@ -27,6 +27,6 @@ pub enum Error {
     /// New version must be greater than the most recent version
     VersionMustBeGreaterThanCurrent = 11,
     /// Invalid name.
-    /// Must be 64 characters or less and be alphanumeric or '_'
+    /// Must be 64 characters or less; ascii alphanumeric or '_';  start with a letter; and not be a Rust keyword
     InvalidName = 12,
 }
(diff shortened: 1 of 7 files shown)


contracts/registry/src/test.rs
```diff
@@ -107,3 +110,31 @@ fn returns_most_recent_version() {
     // let res = client.fetch(name, &None);
     // assert_eq!(res, third_hash);
 }
+
+fn test_string(s: &str, result: bool) {
+    assert!(
+        is_valid(&to_string(s)) == result,
+        "{s} should be {}valid",
+        if !result { "in" } else { "" }
```
chadoh, line 118: cute

contracts/registry/src/error.rs
```diff
@@ -26,4 +26,7 @@ pub enum Error {
     AdminOnly = 10,
     /// New version must be greater than the most recent version
     VersionMustBeGreaterThanCurrent = 11,
+    /// Invalid name.
+    /// Must be 64 characters or less and be alphanumeric or '_'
```
chadoh, line 30: Looks like there's a couple other rules? Should we say them here?

```suggestion
    /// Must be 64 characters or less and be alphanumeric or '_', and start with a letter
```

chadoh: Can we say `and contain only alphanums & '-'`? Is that a thing, "alphanums"?

chadoh (approved):
Wow, string manipulations in soroban 😖

contracts/registry/src/test.rs
willemneal, line 118:
```suggestion
        if result { "" } else { "in" }
```

contracts/registry/src/error.rs
willemneal: Technically it's ascii alphanumeric. And I think it's clearer than "alphanums" as I've never heard that and a quick search turned up nothing but a broken site.

contracts/registry/src/error.rs
```diff
@@ -26,4 +26,7 @@ pub enum Error {
     AdminOnly = 10,
     /// New version must be greater than the most recent version
     VersionMustBeGreaterThanCurrent = 11,
+    /// Invalid name.
+    /// Must be 64 characters or less; ascii alphanumeric or '_';  start with a letter; and not be a Rust keyword
```
willemneal, line 30:
```suggestion
    /// Must be 64 characters or less; ascii alphanumeric or '_'; start with a letter; and not be a Rust keyword
```

PR #71, opened 2025-05-22
feat!: Use semver::Version instead of custom type

willemneal:
Since Cargo.toml stores versions as strings and users will expect to be able to use them directly when passing a version, this PR uses strings as inputs and stores the versions as strings. It also uses the `semver` crate to validate that the input version strings and to ensure that a publish can only publish a version that is greater than the most recent one!

This also has the added benefit of handling any version type cargo supports such as prereleases!


BlaineHeffron (approved):
Looks good but there is a failing check.

PR #80, opened 2025-05-28
fix: GitHub release workflows and extra workflow

pselle:
Removes a commented out workflow and edits the release workflows that are currently erroring on every run, because they have invalid tag patterns.

You don't see failed runs of these CD workflows now, I believe, because they are (correctly) not running on this PR because it is not a tagged release.


willemneal:
Looking at the CI error for Release-Plz we need to add `publish = false` to the new test crate. Would be good to add to this PR since it is CI focused.

willemneal (approved):

PR #86, opened 2025-06-04
feat(registry): Update readme with instructions for mainnet; Update to newest loam_sdk; Use correct contract Ids

willemneal:
- The `stellar-registry-cli` has been updated with instructions for using the registry and how to use it to publish and deploy to mainnet!

- The registry contract now uses the updated version of loam_sdk and adheres to the upgradability SEP by now using `upgrade` in the core subcontract and assumes that this is the default method when upgrading using the registry.

- The contract has been deployed to mainnet, testnet, and futurenet and the correct account id has been updated so that contract ids can be resolved correctly for each network.


pselle:
@willemneal About to read through this, but FYI that it's failing checks!

contracts/registry/src/test.rs
```diff
@@ -129,21 +129,22 @@ fn validate_version() {
...
+        client.try_publish(name, address, bytes, version),
         Err(Ok(Error::VersionMustBeGreaterThanCurrent))
     );
     assert_eq!(
-        client.try_publish(name, address, &bytes, &to_string("0.  0.0"),),
+        client.try_publish(name, address, bytes, &to_string("0.  0.0"),),
         Err(Ok(Error::InvalidVersion))
     );
-    client.publish(name, address, &bytes, &to_string("0.0.1"));
+    client.publish(name, address, bytes, new_version);
```
pselle, line 145: Perhaps a silly question since this was already here -- but wouldn't we expect this publish to succeed? The new version is greater than current

crates/stellar-registry-cli/README.md
```diff
@@ -18,46 +18,57 @@ Deploy a published contract with optional initialization parameters:
...
 
 Options:
-- `DEPLOYED_NAME`: Name of the deployed contract to install
-- `--out-dir (-o)`: Directory to save the contract files
-  - Creates: `<out_dir>/<name>.wasm` and `<out_dir>/contract_id.txt`
+- `CONTRACT_NAME`: Name of the deployed contract to install
 
 ## Configuration
 
+`stellar-cli` povides a way to use a default config for accounts and networks:
```
pselle, line 46:
```suggestion
`stellar-cli` provides a way to use a default config for accounts and networks:
```

crates/stellar-registry-cli/README.md
```diff
@@ -66,17 +77,53 @@ stellar registry deploy \
...
+
+Then can interact with it the contract with `stellar-cli`:
+```bash
+stellar contract invoke --id my-token -- --help
 ```
 
 ## Registry Contract
 
-The registry contract is deployed on testnet at:
+You can install the registry contract like any other:
```
pselle, line 90: What's the reason someone would want to do this? I'm unclear why this instruction is here

crates/stellar-registry-cli/README.md
```diff
@@ -66,17 +77,53 @@ stellar registry deploy \
...
+```bash
+stellar registry install registry
+```
+
+```
+stellar contract invoke --id registry -- --help
+```
+
+### Transitioning to mainnet
```
pselle, line 99:
```suggestion
### Transitioning to Mainnet
```
Maybe we're consistent with mainnet vs Mainnet?

crates/stellar-registry-cli/README.md
```diff
@@ -66,17 +77,53 @@ stellar registry deploy \
...
+
+Once you are satisfied with your contract you can publish and deploy on Mainnet.
+
+The first step is adding mainnet to your `stellar-cli`. [See the reccommend list of RPC provides here]( https://developers.stellar.org/docs/data/rpc/rpc-providers)
```
pselle, line 103:
```suggestion
The first step is adding Mainnet to your `stellar-cli`. [See the reccommend list of RPC provides here]( https://developers.stellar.org/docs/data/rpc/rpc-providers)
```

crates/stellar-registry-cli/README.md
```diff
@@ -66,17 +77,53 @@ stellar registry deploy \
...
 ```
-CC2FLEJRHB2Q5JOAJNPFZU25ZAY6IFYXJL7UQ5GF36F3G5QZ4CQILUID
+or if while in the current directory:
+```bash
+echo STELLAR_NETWORK=mainnet >> .env
 ```
 
+## Publishing and then deploying
+
+Publishing and deploing are exactly the same!
```
pselle, line 124:
```suggestion
Publishing and deploying are exactly the same!
```

Add a code sample here?

pselle: Or do you mean "Publishing and then deploying [on Mainnet] is the same as on any other environment"

contracts/registry/src/test.rs
willemneal: It does! I could make it clearer by using the `try_publish` version and checking `Ok(Ok(())` but in this case the function panics if the method fails. Then the next test tries to publish the old version and it fails.

crates/stellar-registry-cli/README.md
willemneal: To deploy your contract you'll need to publish first but you can also directly deploy already published contracts. Should I make that clearer?

crates/stellar-registry-cli/README.md
willemneal: This is just highlighting that the registry is just a normal contract and that you can interact with it directly if so desired. But mainly it was to highlight the `install` command on something that would work now. Since I still haven't published anything else to the contract yet...

contracts/registry/src/test.rs
pselle: ohhh I see how I misread it, this is an unsaved value that publishes the next version, and the following assert is trying to publish older version. I get it.

crates/stellar-registry-cli/README.md
```diff
@@ -66,17 +77,42 @@ stellar registry deploy \
...
 ```
-CC2FLEJRHB2Q5JOAJNPFZU25ZAY6IFYXJL7UQ5GF36F3G5QZ4CQILUID
+or if while in the current directory:
+```bash
+echo STELLAR_NETWORK=mainnet >> .env
 ```
 
+## Publishing and then deploying
+
+Publishing and deploying are exactly the same!
```
pselle, line 113: Are they? (Responding to fresh review request)

pselle (commented):
Almost there! The deploy/publish are exactly the same comment is still there, and the PR is still failing checks

pselle:
Still failing checks (needs another pass at fmt). I'd be good with ✅ ing this once it's green

Issue #81, opened 2025-05-30
--build-clients should still build clients after account is found


pselle:
When running the scaffold cli with`--build-clients`, it won't build clients if it finds something it's already done, or an error:

```
✅ Build Complete
🌐 using network at https://our-custom-rpc-provider.cool
🔐 creating keys for "official-team-account"
An identity with the name 'official-team-account' already exists
// No JS packages built
```

or running and encountering an error, it's not very clear that there's an error or what do to about it/why clients aren't built

```
🔐 creating keys for "me"
An identity with the name 'me' already exists
📲 installing "soroban_hello_world_contract" wasm bytecode on-chain...
error: Account not found: GCD5LY2VKHIH24ODH7DF2PLT7SJHKPIFFHRHNMRA7ZA2OAUW6FXCN7LF
```

We should either carry on and build the JS packages, and report other errors, or make it clear that an error has occurred and work hasn't happened.

PR #88, opened 2025-06-04
fix: continue building clients on error

BlaineHeffron:
Now when building clients, if there is an error while processing a contract it will report the error and continue to the next contract. A summary of errors is reported at the end. Resolves https://github.com/AhaLabs/scaffold-stellar/issues/81


BlaineHeffron:
> I would collect the results once, then you can partition them. Currently this prints the errors twice.

I found when testing that it was helpful to see where it errored in the execution flow. However, you are correct that it is redundant. I can remove the error print in the loop.

Issue #85, opened 2025-06-04
[bug] registry: publish fails if name is not in metadata


BlaineHeffron:
The contract `publish` function requires a `wasm-name` argument which is passed via the wasm metadata in the cli `publish` command. However, if the wasm was not built with `stellar scaffold`, it will not have this metadata. There should be a way of passing the name via the cli in case its not in the metadata of the wasm.

PR #90, opened 2025-06-05
fix: [registry-cli] add wasm_name arg to publish 

BlaineHeffron:
If there is no contract meta `name`, `publish` command fails due to no `wasm_name` arg. This allows one to specify it via the CLI. Resolves https://github.com/AhaLabs/scaffold-stellar/issues/85

### DIFF
--- a/crates/stellar-registry-cli/src/commands/publish.rs
+++ b/crates/stellar-registry-cli/src/commands/publish.rs
@@ -89,7 +89,7 @@ impl Cmd {
         }
 
         // Add version if specified
-        if let Some(ref version) = self.binver{
+        if let Some(ref version) = self.binver {
             args.push(format!("--version={version}"));
         }
 


crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -67,12 +70,24 @@ impl Cmd {
             ScMetaEntry::ScMetaV0(ScMetaV0 { key, val }) => {
                 let key_str = key.to_string();
                 match key_str.as_str() {
-                    "name" => Some(format!("--wasm_name={val}")),
+                    "name" => {
+                        if self.wasm_name.is_none() {
+                            Some(format!("--wasm_name={val}"))
+                        } else {
+                            None
+                        }
```
willemneal, line 78:
```suggestion
                        
```

BlaineHeffron:
Added binver as well

crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -67,12 +73,21 @@ impl Cmd {
...
+                        .is_none()
+                        .then(|| format!("--wasm_name={val}")),
+                    "binver" => self.binver.is_none().then(|| format!("--version={val}")),
                     _ => None,
                 }
             }
         }));
+
+        // Add wasm_name if specified
+        if let Some(ref wasm_name) = self.wasm_name {
```
pselle, line 87: It looks like you're missing the equivalent block for `Add binver if specified`. Did #85 have the same issue if you had a name in the metadata, but not a binver? If so, you would want to do  as you're doing here.

BlaineHeffron: Good catch, fixed

pselle (approved):

crates/stellar-registry-cli/src/commands/publish.rs
```diff
@@ -67,12 +73,26 @@ impl Cmd {
...
+            args.push(format!("--wasm_name={wasm_name}"));
+        }
+
+        // Add version if specified
+        if let Some(ref version) = self.binver {
```
pselle, line 92: Oops I approved, but should this be ref version or ref binver? Flagging just in case

BlaineHeffron: Its just a temporary variable name so it doesn't matter. I just chose version since the actual contract arg is version but I could have chosen binver. Not sure what is more readable.

willemneal: I usually prefer `.as_ref` on Options. `Option<T>` becomes `Option<&T>`. This does the same. There is also `as_deref` which will call deref too so for example `Option<String>` -> `Option<&str>`. And to add to it you can do `&self.binver`. So in general I prefer the function call. But `ref` predates all so you will se it from time to time.

[See here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=%23%21%5Ballow%28dead_code%29%5D%0A%0Afn+main%28%29+%7B%0A++++let+b+%3D+12%3B%0A++++let+s+%3D+Some%28format%21%28%22hello+%7Bb%7D%22%29%29%3B%0A++++let+Some%28ref+inner%29+%3D+s+else+%7B+todo%21%28%29+%7D%3B%0A++++let+Some%28inner_as_ref_1%29+%3D+s.as_ref%28%29++else+%7B+todo%21%28%29+%7D%3B%0A++++let+Some%28inner_as_ref%29+%3D+%26s++else+%7B+todo%21%28%29+%7D%3B%0A++++let+Some%28inner_as_deref%29+%3D+s.as_deref%28%29++else+%7B+todo%21%28%29+%7D%3B%0A++++assert%21%28inner_as_ref_1+%3D%3D+inner%29%3B%0A++++assert%21%28inner_as_ref+%3D%3D+inner%29%3B%0A++++assert%21%28inner_as_ref.as_str%28%29+%3D%3D+inner_as_deref%29%3B%0A%7D%0A)

PR #106, opened 2025-06-12
fix: ensure account gets funded even if identity already exists

BlaineHeffron:
Currently if you have an identity saved for an account created locally, whenever your network container resets this account will no longer exist. When we go to generate the account, the `generate --fund` call to `stellar-cli` fails with `identity already exists`. Currently we dont exist on this error because it shouldn't cause the remainder of the code execution to fail. However, if the underlying account no longer exists on-chain, any later commands using that account will fail. So to fix it, I added a `keys fund` call if we find the identity exists. This ensures the key actually does exist on-chain and is funded. This is only done when the network is not mainnet, and the environment is `development` or `testing`.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -353,7 +353,7 @@ export default new Client.Client({{
     }
 
     async fn handle_accounts(
-        self,
+        &self,
         accounts: Option<&[env_toml::Account]>,
         network: &Network,
     ) -> Result<(), Error> {
@@ -395,7 +395,6 @@ export default new Client.Client({{
             };
 
             match generate_cmd.run(&args).await {
-                Ok(()) => {}
                 Err(e) if e.to_string().contains("already exists") => {
                     eprintln!("{e}");
                     // Check if account exists on chain
@@ -422,7 +421,7 @@ export default new Client.Client({{
                         fund_cmd.run(&args).await?;
                     }
                 }
-                Err(e) => return Err(e.into()),
+                other_result => other_result?,
             }
         }
 


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -358,18 +366,40 @@ export default new Client.Client({{
...
-                .or_else(|e| {
-                    if e.to_string().contains("already exists") {
-                        // ignore "already exists" errors
-                        eprintln!("{e}");
-                        Ok(())
-                    } else {
-                        Err(e)
+            {
+                Ok(()) => {}
+                Err(e) if e.to_string().contains("already exists") => {
```
pselle, line 379: I was thinking we would only fund if we got an "account doesn't exist" situation.

A flow like:
- create the keys, ignore already exists errors
- check that the account exists
- account doesn't exist? fund on not-mainnet

The approach here _always_ funds if the account currently exists, which would mean that on many random days (since you can fund 1x day) you'd fund your account when you start your app. This seems weird, and undesirable.

BlaineHeffron:
I noticed when changing the fund to within the contract upload error check that the after_deploy script was in the wrong place. It should be within the else statement (if no contract ID is supplied) because it should only be run after actually deploying the contract. You can see in the latest change this was moved to after the deploy call. This fixes after_deploy being run for contracts that have already been deployed and it fixes a potential issue of an account not being funded when it doesn't exist if you have the contract ID already set (all remaining actions are just generating typescript bindings / npm installing none of which require an account).

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -553,19 +565,61 @@ export default new Client.Client({{
...
+                    .into_result()
+                    .expect("no hash returned by 'contract upload'")
+                    .to_string();
+                eprintln!("    ↳ hash: {hash_str}");
+                Ok(hash_str)
+            }
+            Err(e) if e.to_string().contains("Account not found") => {
+                eprintln!("Account not found, attempting to fund...");
+
+                // Get the default account name from environment
```
pselle, line 590: Do we set up this env var when we're running/we'd expect this to work?

pselle (commented):
I like this retry path better, but I'll defer to @willemneal for the final call/what direction this should go.

crates/stellar-scaffold-cli/src/commands/build/clients.rs
BlaineHeffron: Yes, we do. It should always be there - I'll use `expect` instead.

pselle:
I tried installing this and testing it end to end. Steps:

- `cargo install --path .` to install this branch
- created a new project with `stellar scaffold init foo`
- npm i, cp .env, npm start
- stop stellar container `stellar container stop local`
- run npm start again (had to do this twice, but I think that was container related and not us)
- Funding does not work:

```[0] 📲 installing "soroban_hello_world_contract" wasm bytecode on-chain...
[0] Account not found, attempting to fund...
[0] 💸 funding "me"
[0] Build error: Failed to parse JSON from http://localhost:8000/friendbot?addr=GCTMOBHGOR4VVJ5KKSLFNGBKRABVYBYSFNG24AAPLBTLIMUQ7MTYSXLN, trailing characters at line 1 column 5```

File: crates/stellar-scaffold-cli/src/commands/build/clients.rs [hunk 3]
```diff
@@ -358,22 +376,57 @@ export default new Client.Client({{
...
+                        eprintln!("Account not found on chain, funding...");
+                        let fund_cmd = cli::keys::fund::Cmd {
+                            network: Self::get_network_args(network),
+                            address: public_key_cmd,
+                        };
+                        fund_cmd.run(&args).await?;
                     }
-                })?;
+                }
+                Err(e) => return Err(e.into()),
```
willemneal, line 425:
```suggestion
               other_result => other_result?,
```

Issue #105, opened 2025-06-12
Cannot build a project with an OZ added contract


pselle:
Steps to replicate:
- stellar scaffold init foo
- cd foo
- stellar scaffold generate contract --from nft-enumerable
- stellar scaffold build

Expected: Project should build

Actual:  Output errors because folder is nested.
```
error: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `/Users/pselle/Dev/foo/contracts/nft-enumerable`
referenced via `contracts/*` by workspace at `/Users/pselle/Dev/foo/Cargo.toml`

Caused by:
  failed to read `/Users/pselle/Dev/foo/contracts/nft-enumerable/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

NEXT: Unnest the folder:

$ mv contracts/nft-enumerable contracts/foo
$ mv contracts/foo/nft-enumerable contracts/nft-enumerable
$ rm -rf contracts/foo
$ ls contracts
fungible-token-interface hello_world              nft-enumerable
$ stellar scaffold build

Expected: Should be able to build project
Actual: Cannot build project because dependencies are not satisfied

```
error: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `/Users/pselle/Dev/foo/contracts/nft-enumerable`
referenced via `contracts/*` by workspace at `/Users/pselle/Dev/foo/Cargo.toml`

Caused by:
  failed to parse manifest at `/Users/pselle/Dev/foo/contracts/nft-enumerable/Cargo.toml`

Caused by:
  error inheriting `stellar-default-impl-macro` from workspace root manifest's `workspace.dependencies.stellar-default-impl-macro`

Caused by:
  `dependency.stellar-default-impl-macro` was not found in `workspace.dependencies`

```

BlaineHeffron:
Sorry about that, I encountered this issue when testing and thought I had fixed it 🤦

PR #107, opened 2025-06-12
fix: only copy contents to output location

BlaineHeffron:
resolves https://github.com/AhaLabs/scaffold-stellar/issues/105

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs
+++ b/crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs
@@ -421,7 +421,9 @@ fn open_wizard() -> Result<(), Error> {
     eprintln!("   3. Extract the downloaded ZIP file");
     eprintln!("   4. Move the contract folder to your contracts/ directory");
     eprintln!("   5. Add the contract to your workspace Cargo.toml if needed");
-    eprintln!("   6. You may need to modify your environments.toml file to add constructor arguments");
+    eprintln!(
+        "   6. You may need to modify your environments.toml file to add constructor arguments"
+    );
     eprintln!(
         "\n💡 The wizard will generate a complete Soroban contract with your selected features!"
     );
--- a/crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs
+++ b/crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs
@@ -1,7 +1,9 @@
 use clap::Parser;
+use flate2::read::GzDecoder;
 use reqwest;
 use serde::Deserialize;
 use std::{fs, path::Path};
+use tar::Archive;
 
 #[derive(Deserialize)]
 struct Release {
@@ -33,6 +35,12 @@ pub enum Error {
     Io(#[from] std::io::Error),
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
+    #[error(transparent)]
+    CargoToml(#[from] cargo_toml::Error),
+    #[error(transparent)]
+    TomlDeserialize(#[from] toml::de::Error),
+    #[error(transparent)]
+    TomlSerialize(#[from] toml::ser::Error),
     #[error("Git command failed: {0}")]
     GitCloneFailed(String),
     #[error("Example '{0}' not found in OpenZeppelin stellar-contracts")]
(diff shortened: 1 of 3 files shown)


BlaineHeffron:
It doesn't with some of the contracts. They have dependencies from within the repo that need to be copied in as well. I will work on this dependency resolution code.

pselle (commented):
This builds now! Let's add one more bit of helper text, because users might need to add  constructor args, but we won't do it for them (the last line is an addition)

```
$ stellar scaffold generate contract --from nft-enumerable
🔍 Downloading example 'nft-enumerable'...
📂 Using cached repository (release v0.2.0)...
📦 Added the following dependencies to workspace:
   • stellar-default-impl-macro
   • stellar-non-fungible
✅ Successfully downloaded example 'nft-enumerable' to contracts/nft-enumerable
💡 You may need to modify your environments.toml to add constructor arguments!
```

pselle:
Nope, still fails when I try a different example (it worked with nft-enumerable). Try

- `stellar scaffold generate contract --from fungible-token-interface`

I did this is a freshly scaffold'ed project.

Expected: successful build

Actual:
```
$ stellar scaffold build --build-clients
ℹ️  CARGO_BUILD_RUSTFLAGS=--remap-path-prefix=/Users/pselle/.cargo/registry/src= cargo rustc --manifest-path=contracts/fungible-token-interface/Cargo.toml --crate-type=cdylib --target=wasm32v1-none --release
   Compiling fungible-token-interface-example v0.0.1 (/Users/pselle/Dev/scratch/bar/contracts/fungible-token-interface)
error[E0432]: unresolved import `stellar_fungible::Base`
  --> contracts/fungible-token-interface/src/contract.rs:20:5
   |
20 | use stellar_fungible::Base;
   |     ^^^^^^^^^^^^^^^^^^^^^^ no `Base` in the root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `fungible-token-interface-example` (lib) due to 1 previous error
error: exit status exit status: 101

**Comment by BlaineHeffron** [comment]
That was in a standalone but I've tested both.

File: crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs [hunk 1]
```diff
@@ -112,14 +77,138 @@ impl Cmd {
...
+                        if existing_tag != tag {
+                            workspace_deps.insert(
+                                dep.clone(),
+                                cargo_toml::Dependency::Detailed(Box::new(git_dep)),
+                            );
+                            updated_deps.push((dep, existing_tag.clone()));
+                        }
+                    }
+                }
+            } else {
```
willemneal, line 157: Since we are using the newest version of rust this should work.
```suggestion
            if let Some(existing_dep) = workspace_deps.clone().get(&dep) {
                // Check if we need to update the tag
                if let Detailed(DependencyDetail { tag: Some(existing_tag), ..}) = &existing_dep && existing_tag != tag {
                      workspace_deps.insert(
                          dep.clone(),
                          cargo_toml::Dependency::Detailed(Box::new(git_dep)),
                      );
                      updated_deps.push((dep, existing_tag.clone()));
                }
            } else {
```

crates/stellar-scaffold-cli/src/commands/generate/contract/mod.rs
```diff
@@ -112,14 +77,134 @@ impl Cmd {
...
+                    cargo_toml::Dependency::Detailed(detail) => {
+                        if detail.inherited || detail.git.is_some() {
+                            deps.push(dep_name.clone());
+                        }
+                    }
+                }
+            }
+        }
+
+        Ok(deps)
```
willemneal, line 205: I might go over board with my distaste for for loops + mutable vector pushes. But IMO it reads better to use iterators and make it clearer what the goals are.

Also this is definitely not formatted correctly! Sorry!

```suggestion
        Ok(manifest.dependencies
              .iter()
             .filter(|(dep_name, _)| dep_name.starts_with("stellar-"))
             .filter_map((|(dep_name, dep_detail)| {
                if let Dependency::Detailed(detail) && !(detail.inherited || detail.git.is_some()) { 
                    None 
               } else {
                    Some(dep_name.clone())
              }).collect())
```

BlaineHeffron:
@willemneal I tried the suggestion for simplifying the existing workspace deps check but I get "`let` expressions in this position are unstable", reverting

PR #108, opened 2025-06-16
feat: add Cargo metadata

ifropc:
This PR adds support of `[package.metadata.stellar]` section in `Cargo.toml` file.
* `rsver` and `rssdkver` are ignored
* When `cargo_inherit` is set, common Cargo.toml properties will be automatically set (`name`, `authors`, `homepage`, `repository` and `version` -- renamed to `binver` to comply with SEP-47)
* All other keys do not hold any specific restrictions and will be converted to metadata keys following the rules:
* int, bool, string are used as metadata value, with key derived from the parent
* Array of int/bool/strings are folded with ',' as a separator (to reduce WASM size)
* Complex arrays are using cfg files notation (key[0], key[1] and so on) and recursively converted using parent's key suffixed with [0]/[1]/and so on
* Objects are recursively converted, using object's key as metadata key and object's value as metadata value. For nested objects `.` is used as a key separator

Here's an example of toml section:
```toml
[package.metadata.stellar]
rsver = "ignored"
rssdkver = "ignored"
cargo_inherit = true
name = "soroban-registry"
home_domain = "ahalabs.dev" # can be defined either as `homepage` or `home_domain`
repository = "https://github.com/AhaLabs/scaffold-stellar" # will be renamed to source_repo
my_custom_string = "foo"
my_custom_bool = true
my_custom_int = 1

[package.metadata.stellar.object]
simple_array = ["a", 123, false]
complex_array = ["a", { inner_array = ["b", "c"] }]
my_custom_object = { field1 = "val1", field2 = [ "single element array" ] }
```

And how it's converted to metadata (using `stellar contract info meta` to show the end result)
```
> stellar contract info meta --wasm target/stellar/registry.wasm
ℹ️ Loading contract spec from file...
Contract meta:
 • rsver: 1.86.0 (Rust version)
 • rssdkver: 22.0.4#1d4afb3b981a4f4b2bbc19f0ce38af85c4ab316a (Soroban SDK version and it's commit hash)
 • authors: Aha Labs <hello@ahalabs.dev>
 • binver: 0.1.0
 • home_domain: ahalabs.dev
 • my_custom_bool: true
 • my_custom_int: 1
 • my_custom_string: foo
 • name: soroban-registry
 • object.complex_array[0]: a
 • object.complex_array[1].inner_array: b,c
 • object.my_custom_object.field1: val1
 • object.my_custom_object.field2: single element array
 • object.simple_array: a,123,false
 • source_repo: https://github.com/AhaLabs/scaffold-stellar

```

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/mod.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/mod.rs
@@ -213,7 +213,7 @@ impl Command {
                         }
                     }
 
-                    Self::rec_add_meta("".to_string(), &mut meta_map, val);
+                    Self::rec_add_meta(String::new(), &mut meta_map, val);
 
                     // Reserved keys
                     meta_map.remove("rsver");
@@ -237,7 +237,7 @@ impl Command {
             }
         }
 
-        return Ok(cmd);
+        Ok(cmd)
     }
 
     fn rec_add_meta(prefix: String, meta_map: &mut BTreeMap<String, String>, value: &Value) {
@@ -253,7 +253,7 @@ impl Command {
                 meta_map.insert(prefix, s.clone());
             }
             Value::Array(array) => {
-                if array.iter().all(|x| Self::is_simple(x)) {
+                if array.iter().all(Self::is_simple) {
                     let s = array
                         .iter()
                         .map(|x| match x {
@@ -265,7 +265,7 @@ impl Command {
                     meta_map.insert(prefix, s);
                 } else {
                     for (pos, e) in array.iter().enumerate() {
-                        Self::rec_add_meta(format!("{prefix}[{pos}]"), meta_map, e)
+                        Self::rec_add_meta(format!("{prefix}[{pos}]"), meta_map, e);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


crates/stellar-scaffold-cli/README.md
```diff
@@ -82,15 +82,18 @@ The build process ensures:
...
+Contract metadata is set when running `stellar scaffold build`. You can configure metadata with 
+`[package.metadata.stellar]` section in your `Cargo.toml` file.
+For example:
+```toml
+[package.metadata.stellar]
+# When set to `true` will copy over [package] section's `name`, `authors`, `homepage`, `repository` and `version` (renamed to `binver` to comply with SEP-47)
+cargo_inherit = true
+# Override one of the inherited values
+name = "my-awesome-contract"
+homepage = "ahalabs.dev"
```
pselle, line 94: Hmm, I like the inherit, but I thought moving things to our own section means we'd use our own key names, rather than mapping them. meaning we would use source_repo, home_domain, binver… That the names are different is of course an upstream unpleasantness, what do you think about this approach?

pselle: I did the renaming in the implementation because of using [package] keys … I would think if we're moving to our own metadata section, we don't need to rename

pselle (approved):
Based on the description and my reading, this looks good. I would love some unit tests, but won't block on it.

pselle:
@ifropc I approved checks to run on this PR, and it looks like there's a failure/something to fix (usually fmt). Please fix! Also updated your access so you should be able to merge when you're ready.

PR #124, opened 2025-06-26
feat: use global args and printer in build / watch

BlaineHeffron:
Refactors to use `stellar-cli` Print struct for relaying information from `build` and `watch` which respects the `--quiet` flag, passes global args to the CLI commands.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -8,6 +8,7 @@ use std::fmt::Debug;
 use std::hash::Hash;
 use std::process::Command;
 use stellar_cli::commands::NetworkRunnable;
+use stellar_cli::print::Print;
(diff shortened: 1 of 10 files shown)


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -8,6 +8,7 @@ use std::fmt::Debug;
 use std::hash::Hash;
 use std::process::Command;
 use stellar_cli::commands::NetworkRunnable;
+use stellar_cli::print::Print;
 use stellar_cli::utils::contract_hash;
 use stellar_cli::{commands as cli, CommandParser};
```
willemneal, line 13: these can be collapsed into `sellar_cli::{...`

willemneal:
I would remove the printer instance method not pass it and then do something like:
```rust
Print::new(self.global_args.as_ref().map(|args|args.quiet).unwrap_or_default())
```

PR #126, opened 2025-07-02
Fix: Build clients race condition

zachfedor:
Previously, the client build process deleted the old contract code before building and installing the new ones. While running in watch mode concurrently with the frontend dev server, this caused a hot-reload which would throw an error since app code is importing contracts that don't exist until rebuilt.

This change builds and installs the contracts in a temp directory then replaces the contract code at once, ensuring only one hot-reload and no broken imports. This has the added benefit of being able to use the old contract code as a backup if the build/install process fails for any reason.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -359,6 +359,7 @@ export default new Client.Client({{
             }
             eprintln!("✅ Client {name:?} updated successfully");
         } else {
+            std::fs::create_dir_all(&final_output_dir)?;
             // No existing directory, just move temp to final location
             std::fs::rename(&temp_dir, &final_output_dir)?;
             eprintln!("✅ Client {name:?} created successfully");
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -6,6 +6,7 @@ use serde_json;
 use shlex::split;
 use std::fmt::Debug;
 use std::hash::Hash;
+use std::path::Path;
 use std::process::Command;
 use stellar_cli::commands::NetworkRunnable;
 use stellar_cli::utils::contract_hash;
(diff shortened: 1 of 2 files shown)


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -274,30 +287,34 @@ export default new Client.Client({{
...
+
         cli::contract::bindings::typescript::Cmd::parse_arg_vec(&[
             "--contract-id",
             contract_id,
             "--output-dir",
-            output_dir
-                .to_str()
-                .expect("we do not support non-utf8 paths"),
-            "--overwrite",
+            temp_dir.to_str().expect("we do not support non-utf8 paths"),
```
BlaineHeffron, line 317: This produces output like 
```
✅ Generated!
ℹ️ Run "npm install && npm run build" in "/home/blaine/projects/smart-contracts/test6/packages/.temp_fungible_token_interface_example_294186" to build the JavaScript NPM package.
``` 
Perhaps we should suppress this part of the output so that it doesnt confuse the user (since we are going to run npm build in it anyway). Also a question: is this contract bindings typescript call what is deleting the contract client code which is causing the reload?

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -307,48 +324,82 @@ export default new Client.Client({{
         .await?;
 
         eprintln!("🍽️ importing {name:?} contract");
-        self.write_contract_template(name, contract_id)?;
+        // Write contract template to temp directory first
+        self.write_contract_template_to_dir(name, contract_id, &temp_dir)?;
```
BlaineHeffron, line 328: I don't think this is needed; why not just write the contract client after successfully transferring the underlying typescript bindings to the proper directory? I notice that this call creates an extra folder / file in `packages/<name>/src/contracts/<name>.ts. This file should only be in `src/contracts/<name>.ts` it is what imports the typescript bindings that live in `packages/<name>` and creates the contract client for them. (Note that usually `packages/<name>/src` should only contain `index.ts`.)

BlaineHeffron (changes requested):
Nice work, I think this a good solution. You shouldn't need to copy the contract template to the temporary directory since that lives in a separate place from the typescript bindings. I would just remove that (and that should remove the need for the `write_contract_template_to_dir` refactor).

crates/stellar-scaffold-cli/src/commands/build/clients.rs
BlaineHeffron: Nevermind about my question; I see that when overwrite is specified the first thing it does is delete the code which must be what is causing the hot reload -> error.

willemneal (approved):

PR #136, opened 2025-07-08
chore: add/update NFT token example on init

### LINKED ISSUE #133: Missing NFT contract in the contracts directory
It's okay to have hello world which is very simple. But we should minimally have the FT and NFT OpenZeppelin contract in this repo from the get go. This is described in the roadmap milestone. There could be three contracts for developers

- hello world
- fungible token contract
- nonfungible token contract

pselle:
Bases off of #127 , so this can be rebased if/when that's merged, or merge this instead.

Refactor the updating of the fungible token to also update the nft example.

Closes #133


BlaineHeffron (approved):
Looks good, I think it just needs to be rebased to fix the merge conflict

PR #149, opened 2025-07-18
feat: `watch` command calls upgrade instead of redeploying when possible

ifropc:
`watch` command now calls an `upgrade` method on a contract instead of redeploying.
It must satisfy following criteria:
1. Contract is already deployed (has an alias)
2. Existing contract has `upgrade` function that follows SEP-49
3. New WASM hash also has an `upgrade` function that follows SEP-49
Tested manually with openzeppelin upgreable contract

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::struct_excessive_bools)]
+use super::env_toml::Network;
 use crate::commands::build::env_toml;
 use indexmap::IndexMap;
 use regex::Regex;
@@ -12,9 +13,8 @@ use stellar_cli::{
     commands as cli, commands::NetworkRunnable, print::Print, utils::contract_hash, CommandParser,
 };
 use stellar_strkey::{self, Contract};
-use stellar_xdr::curr::Error as xdrError;
-
-use super::env_toml::Network;
+use stellar_xdr::curr::ScSpecEntry::FunctionV0;
+use stellar_xdr::curr::{Error as xdrError, ScSpecEntry, ScSpecTypeBytesN, ScSpecTypeDef};
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum)]
 pub enum ScaffoldEnv {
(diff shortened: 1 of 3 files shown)


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -792,6 +810,87 @@ export default new Client.Client({{
...
+        let legacy_upgradeable = Self::is_legacy_upgradeable(existing_spec);
+
+        if legacy_upgradeable.is_none() {
+            return Ok(None);
+        }
+
+        if Self::is_legacy_upgradeable(spec_to_upgrade).is_none() {
+            printer.warnln("New WASM is not upgradable. Contract will be redeployed instead of being upgraded.");
+            return Ok(None);
+        }
```
willemneal, line 843:
```suggestion
        if !Self::is_legacy_upgradeable(existing_spec) {
            return Ok(None);
        }

        if !Self::is_legacy_upgradeable(spec_to_upgrade) {
            printer.warnln("New WASM is not upgradable. Contract will be redeployed instead of being upgraded.");
            return Ok(None);
        }
```

ifropc: `is_legacy_upgradeable` returns an option, it's `None` if contract is not upgradable, `Some(true)` if it's legacy upgradable (to be compatible with current version of the SEP), and `Some(false)` if it's newer upgradable contract (so it doesn't have an operator argument)
We can drop support of legacy in the future (but for now I think it's better to keep it so scaffold works on upgradable OZ contracts)

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -792,6 +810,87 @@ export default new Client.Client({{
...
+        let mut redeploy_args = vec![
+            "--id".to_string(),
+            existing_contract_id.to_string(),
+            "--".to_string(),
+            "upgrade".to_string(),
+            "--new_wasm_hash".to_string(),
+            hash.to_string(),
+        ];
+
+        if legacy_upgradeable.is_some_and(|x| x) {
```
willemneal, line 857:
```suggestion
        if legacy_upgradeable {
```

crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -792,6 +810,87 @@ export default new Client.Client({{
...
+            .collect();
+        cli::contract::invoke::Cmd::parse_arg_vec(&redeploy_args)?
+            .run_against_rpc_server(self.global_args.as_ref(), None)
+            .await?
+            .into_result()
+            .expect("no result returned by 'contract invoke'");
+        printer.infoln(format!(
+            "Contract upgraded: {}",
+            existing_contract_id.to_string()
+        ));
```
willemneal, line 876:
```suggestion
        printer.infoln(format!("Contract upgraded: {existing_contract_id}"));
```

PR #175, opened 2025-08-28
feat(registry): ensure that all names are converted to kebab case

willemneal:
This way `crate_name` becomes `crate-name`, ensuring that there is no way for another `crate-name` to be published or deployed.


chadoh (changes requested):
Tests?

willemneal:
Yeah hence the draft. Been working on tests for the other PR.

PR #180, opened 2025-09-03
fix(build-clients): only `allowHttp` in dev & test

chadoh:
This got switched around by mistake in 1bb8c7e35af9b5632f0515aab5c5c929fe58a870

We need tests on this! But for now it's a show-stopping bug (no local environment works correctly without this), and I have verified that it works correctly by using it in a local project.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -262,7 +262,9 @@ impl Args {
     }
 
     fn create_contract_template(&self, name: &str, contract_id: &str) -> Result<(), Error> {
-        let allow_http = if self.stellar_scaffold_env(ScaffoldEnv::Production) == "production" {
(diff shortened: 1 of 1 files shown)


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -262,7 +262,9 @@ impl Args {
     }
 
     fn create_contract_template(&self, name: &str, contract_id: &str) -> Result<(), Error> {
-        let allow_http = if self.stellar_scaffold_env(ScaffoldEnv::Production) == "production" {
+        let allow_http = if ["development", "test"]
```
willemneal, line 265: raw strings ;-(

willemneal (approved):

crates/stellar-scaffold-cli/src/commands/build/clients.rs
chadoh: love the frowny wink

PR #193, opened 2025-09-23
feat: first draft of tutorial

zachfedor:
TODOs:

- [x] complete 01-getting-started with build and deploy to local net
- [x] add running UI to first section of 02-making-improvements and show Contract Explorer
    - this is a good time to show the new private function doesn't show up
    - side by side w/ CLI
    - invoke via alias
- [x] changing above contract function requires changing react component
- [ ] finish final step of obfuscating the number

### DIFF
--- a/website/docs/tutorial/03-adding-payments.md
+++ b/website/docs/tutorial/03-adding-payments.md
@@ -76,18 +76,18 @@ pub fn guess(env: &Env, guesser: Address, a_number: u64) -> bool {
 
       // pay full pot to `guesser`, whether they sent the transaction or not
       let tx = xlm_client.transfer(
-        self.current_contract_address(),
+        env.current_contract_address(),
         guesser,
-        x.balance(self.current_contract_address()),
+        xlm_client.balance(env.current_contract_address()),
       );
-      if tx.is_err {
+      if tx.is_err() {
         panic!("transfer failed!");
       }
   } else {
     // Before transferring their funds, make sure guesser is actually the one calling this function
     guesser.require_auth();
-    let tx = xlm_client.transfer(guesser, self.current_contract_address(), 1_000_000_0);
-    if tx.is_err {
+    let tx = xlm_client.transfer(guesser, env.current_contract_address(), 1_000_000_0);
+    if tx.is_err() {
       panic!("transfer failed!");
     }
   }


docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+And here is the reset function. Note that we use `require_admin()` here so only you can run this function. It generates a random number between 1 and 10 and uses our key to store it.
+
+```rust
+    /// Guess a number between 1 and 10
+    pub fn guess(env: &Env, a_number: u64) -> bool {
+        a_number == env.storage().instance().get::<_, u64>(&THE_NUMBER).unwrap()
+    }
+}
```
pselle, line 138: Because of the contracts refactor, you can reference public/private functions if you like. Not sure if it's necessary, but it's an option

docs/tutorial/03-adding-in-transactions.md
```diff
@@ -0,0 +1,508 @@
...
+```
+
+The `token` import gives us access to Stellar's built-in token functionality.
+
+## Step 2: 💰 Add Economic Storage Variables
+
+Let's add storage for tracking our economic state. Add these constants:
+
+```rust
+const THE_NUMBER: Symbol = symbol_short!("n");
```
pselle, line 45: Might want to move this to a different sort of storage structure … https://developers.stellar.org/docs/build/smart-contracts/example-contracts/custom-types#how-it-works @chadoh ?

docs/tutorial/03-adding-in-transactions.md
```diff
@@ -0,0 +1,508 @@
...
+const GUESS_FEE: Symbol = symbol_short!("fee");        // Cost per guess
+const NATIVE_TOKEN: Symbol = symbol_short!("native");   // XLM token address
+```
+
+## Step 3: Update the Constructor
+
+Let's enhance our constructor to set up the economic parameters:
+
+```rust
+pub fn __constructor(env: &Env, admin: &Address, initial_pot: u64, guess_fee: u64) {
```
pselle, line 57: I think you mean to remove this initial_pot argument, since later you say the pot starts empty

docs/tutorial/03-adding-in-transactions.md
```diff
@@ -0,0 +1,508 @@
...
+````
+
+## Step 8: Test the Economic System
+
+Now let's test our new economic features:
+
+### Build and Deploy
+
+```bash
+stellar contract build
```
pselle, line 336: Again, should use stellar scaffold tooling here

docs/tutorial/03-adding-in-transactions.md
```diff
@@ -0,0 +1,508 @@
...
+
+You should see `10000000` (1 XLM)!
+
+### Make Some Paid Guesses
+
+Now let's have different users make guesses:
+
+```bash
+# Bob makes a guess (and pays the fee)
+stellar contract invoke \
```
pselle, line 378: Maybe we can leverage aliases in the tutorial so we can write this id here?

docs/tutorial/03-adding-in-transactions.md
```diff
@@ -0,0 +1,508 @@
...
+  -- guess \
+  --guesser $(stellar keys address bob) \
+  --a_number 7
+```
+
+When someone guesses correctly, they'll receive the entire pot, and the pot will reset to 0!
+
+## Understanding the Economic Flow
+
+Let's trace through what happens:
```
pselle, line 415: This would be a great spot for a graphic

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,357 @@
...
+- compiling
+- debugging
+- testing
+- deploying
+- and upgrading
+
+
+## ✅ Prerequisites
+
+You should have a basic understanding of using the command line and of general programming concepts. Stellar Contracts are written in a subset of Rust, although we'll walk through the code together so don't worry if this is your first time with the language.
```
pselle: "Stellar smart contracts"

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,357 @@
...
+    • set_admin
+    • upgrade
+✅ Build Complete
+```
+
+We implemented a few of those functions in our contract (i.e. `__constructor`, `guess`, and `reset`), but where did the others come from? They are inherited from the `Administratable` and `Ugradable` traits we talked about earlier.
+
+## 🚢 Deploy To Local Network
+
+In order to deploy the contract, we need an account on the local network to own the contract and sign transactions. Let's do that now and also provide some [fake funds](https://developers.stellar.org/docs/learn/fundamentals/networks#friendbot) to test with.
```
pselle: This should use scaffold stellar tooling. Scaffold stellar build should have made the keys for the local network.

pselle (changes requested):

docs/tutorial/01-getting-started.md
chadoh: Echoing Pam, but also suggesting that we don't say "a subset of" Rust. It's Rust without its standard library, and with some interesting libraries like soroban-sdk instead, but it's still "just Rust."

```suggestion
You should have a basic understanding of using the command line and of general programming concepts. Stellar smart contracts are written in Rust, although we'll walk through the code together so don't worry if this is your first time with the language.
```

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,357 @@
...
+
+TODO: rewrite our own?
+[Follow original setup instructions.](https://developers.stellar.org/docs/build/smart-contracts/getting-started/setup)
+
+
+## 🏗️ Create the Scaffold
+
+Our smart contract will be a Guess The Number game. You (the admin) can deploy the contract, randomly select a number between 1 and 10, and seed the contract with a prize. Users can make guesses and win the prize if they're correct!
+
+> ℹ️ Why a guessing game? A standard Hello World program isn't all that useful of an example. Unlike most smart contracts, there's nothing to interact with. That's why the Rust Programming Language book uses a guessing game as [their first tutorial project](https://doc.rust-lang.org/book/ch02-00-guessing-game-tutorial.html). We thought we'd do the same.
```
chadoh: I assume we're using Docusaurus, to match main Stellar docs? If so, we can use an [admonition box](https://docusaurus.io/docs/2.x/markdown-features/admonitions) (weird name for it, imho):

```suggestion
:::info Why a guessing game? 

A standard Hello World program isn't all that useful of an example. Unlike most smart contracts, there's nothing to interact with. That's why the Rust Programming Language book uses a guessing game as [their first tutorial project](https://doc.rust-lang.org/book/ch02-00-guessing-game-tutorial.html). We thought we'd do the same.

:::
```

Although, honestly, I wonder if this info (er, "admonition") box is worth having at all. If I were going through this tutorial, I would not be wondering why it's not guiding me through a "standard Hello World program." I'd just be like, "ok, cool/whatever, a guessing game, sounds interesting enough."

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,357 @@
...
+
+Let's use the Stellar CLI tool to get a starting point. Open your terminal and navigate to the directory you keep your projects, then type:
+
+```bash
+$ stellar scaffold init my-project
```
chadoh: I _think_ we want to keep the `$` out of it, so people can copy-paste correctly. I think I've seen total newbies get confused by the dollar sign.

chadoh (commented):
A bundle of comments!

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,367 @@
...
+pub struct GuessTheNumber;
+```
+
+The `#[...]` syntax in Rust is called an [attribute](https://doc.rust-lang.org/reference/attributes.html). It's a way to label code for the compiler to handle it with special instructions. *Inner* attributes (with the `#!`) apply to the scope they're within (meaning `!#[no_std]` applies to the whole file/module), and *Outer* attributes (just the `#`) apply to the next line.
+
+Here we're defining a [struct](https://doc.rust-lang.org/book/ch05-01-defining-structs.html) (a "structure" to hold values) and applying attributes of a Stellar smart contract.
+
+```rust
+const THE_NUMBER: Symbol = symbol_short!("n");
+pub const ADMIN_KEY: &Symbol = &symbol_short!("ADMIN");
```
chadoh: For later: let's circle back and convert this to use a `pub enum DataKey` a la [liquidity_pool](https://github.com/stellar/soroban-examples/blob/adcff845ca10aebe7aa4fc24ec8bd4ea09681ee1/liquidity_pool/src/lib.rs#L8-L17). This is a more standard way to handle multiple storage keys in Soroban contracts. Willem wanted to switch away from this when he and I first created this GuessTheNumber contract because using a whole enum for a single key feels silly. But now we don't have a single key! And designing the contract with more standard patterns in a way that grows in complexity more easily seems like a good thing.

When we finally have `contracttrait` and can move the `ADMIN` key back out of this contract, we can revisit this. Might make sense to go back to `symbol_short` here, then introduce `enum DataKey` in a later step of the tutorial when we add more data to the contract.

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,367 @@
...
+### Docker
+
+We'll run a local Stellar network inside a Docker container, so head to the [Get Docker page](https://docs.docker.com/get-started/get-docker/) and follow the instructions for installing Docker Desktop for your operating system. Once it's installed, open it up. It needs to be running in the background but then Scaffold Stellar will handle the rest.
+
+### Scaffold Stellar
+
+Lastly, we'll install the Scaffold Stellar plugin for the Stellar CLI.
+
+```bash
+cargo install --locked stellar-scaffold-cli
```
willemneal: Should we also suggest cargo-binstall? It would help speed up the install.

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,367 @@
...
+Connect wallet to play the guessing game
+```
+
+In order to test out our deployed example contract, we'll need to connect to a wallet.
+
+### 💰 Connect a Wallet
+
+In the top right corner, you'll see a big "Connect" button. Click it. You need to have a [Wallet](https://stellar.org/learn/wallets-to-store-send-and-receive-lumens) in order to interact with the dApp. The modal that opened will show a few options if you don't have one already. We recommend using [Freighter](https://www.freighter.app/).
+
+Once it's installed, we need to connect it to our local network running in Docker. Open the extension, click the menu, and navigate to "Settings," then "Network." Click the "Add custom network" button and enter the following info:
```
willemneal: Outside the scope of this, but it would  be *great* if freighter could have these defaults.

docs/tutorial/01-getting-started.md
willemneal: While I always prefer `&` here since you always refer to them as references, I say we stick with one or the other. Perhaps the value as to not introduce references at this point.

docs/tutorial/01-getting-started.md
```diff
@@ -0,0 +1,367 @@
...
+[0]     • reset
+[0]     • upgrade
+[0] ✅ Build Complete
+```
+
+You'll see a build summary for each of our contracts as they compile to Wasm, the code that will actually run on the blockchain. There's a few other functions in our contract that weren't exported here. That's because they're private. We'll implement our own in the next part of the tutorial and explain why they're useful.
+
+```
+[0] ℹ️ Creating keys for "me"
+[0] ✅ Key saved with alias me in "/Users/zach/code/guessing-game-tutorial/.config/stellar/identity/me.toml"
```
willemneal: I updated so that we use the global config to ensure that `me` can be found outside the project and other identities can also be found in the project.

docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+
+```bash
+$ npm start
+```
+
+In the header menu, you should see a `</> Debugger` link. This will open up our handy dandy Contract Explorer. This will let you explore all the contracts within your project, view the documentation for every method, and even invoke them with arguments directly from the UI!
+
+This is a great tool to help debug and test while you develop. So give it a shot! Navigate to the debugger, select the `guess_the_number` contract, and test out our new `reset` method by hitting it's `Submit` button.
+
+Our `reset` method is available to be called by code _outside_ our contract because we opted in to it being a public method with the `pub`. Our `set_random_number` is private by default, it's not even visible to the outside world. It's not listed in the Contract Explorer. It's not listed in the CLI help either:
```
willemneal:
```suggestion
Our `reset` method is available to be called by code _outside_ our contract because we opted in to it being a public method with the `pub` keyword. Our `set_random_number` is private by default, it's not visible to the outside world. It's not listed in the Contract Explorer. It's not listed in the CLI help either:
```

docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+### Understanding `expect()` vs `unwrap()`
+
+Because [Rust distinguishes between two types of errors](https://doc.rust-lang.org/book/ch09-00-error-handling.html), there's two different ways to handle them. Some errors are from bugs, and obviously we want to avoid these as much as possible. But some errors are expected. And we can recover from them without crashing. Both of these methods are shortcuts to check for values and error, but one gives you the ability to explain _why_:
+
+- `unwrap()`: Crashes with a generic error message
+- `expect()`: Crashes with your custom error message
+
+Both will panic if the value is `None`, but `expect()` gives better debugging info. In our case, this should never happen since we now set a number in the constructor, but it's good defensive programming. For example, we were already using it in our `require_admin` method for authentication.
```
willemneal: Hmm I need to check but I thought that message provided by expect are not surfaced when calling the contract. If so, perhaps best to leave this until we introduce errors or introduce them now?

docs/tutorial/03-adding-in-transactions.md
chadoh: good 'nuff for  now tbh

chadoh (commented):
it looks good through Step 2, and Step 3 has a good start. I'd like to see the TODO items in Step 3 cleaned up a bit (like, the headings jump from Step 4 to Step 7 for some reason), but then I think we could merge as-is. We don't need to finish it to merge, but the "TODOs" and in-progress nature of it should make sense, and potentially help a motivated hackathoner send us improvements where they expand stub sections and implement our TODO items.

website/docs/tutorial/00-overview.md
```diff
@@ -0,0 +1,20 @@
...
+---
+# Tutorial Overview
+
+This tutorial will help you learn how to use [Scaffold Stellar](https://github.com/theahaco/scaffold-stellar) to build and manage smart contracts on the Stellar blockchain and a decentralized application (dApp) to interact with them. Scaffold Stellar is a developer toolkit that provides CLI tools, contract templates, and a starter React UI to get your idea out of your head and on to the network as fast as possible.
+
+:::tip
+If you just want to get up and running quickly, check out the [Quick Start](/docs/quick-start.mdx) guide.
+:::
+
+Before jumping in, you should have a basic understanding of the command line and of general programming concepts, but we'll walk through all the code together so don't worry if you're new to Stellar, Rust, or dApp development. We'll link out to [The Rust Programming Language book](https://doc.rust-lang.org/stable/book/) to explain concepts if you want to dive deeper.
```
tupui: We could also link to the official docs here and recommend people to take some time there https://developers.stellar.org

website/docs/tutorial/01-getting-started.md
tupui: I would have put optimisations like this in the improvement section otherwise we introduce many concepts at once.

website/docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,501 @@
...
+reset
+"""
+```
+
+This is a [Toml table](https://toml.io/en/v1.0.0-rc.2#table). See the TOML spec for other ways you could specify the same information.
+
+Let's walk through this line by line:
+
+- `[development.contracts.guess_the_number]`: this project only has one contract, so we can specify the settings for its contract clients here. You could also have a `[development.contracts]` with a more JSON-like specification for `guess_the_number` (`guess_the_number = { client = true, … }`).
+- `guess_the_number`: this name must match the name of the contract specified in its `Cargo.toml` file, but in underscore-case. Compare it to the `name` field in `contracts/guess-the-number/Cargo.toml` and the generated Wasm files (`ls target/wasm32v1-none/release/*.wasm`).
```
tupui: > but in underscore-case

Note: sounds like we should take care of that ourselves.

website/docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,501 @@
...
+✅ Key saved with alias bob in ".config/stellar/identity/bob.toml"
+✅ Account bob funded on "Standalone Network ; February 2017"
+
+$ stellar contract invoke --id guess_the_number --source bob --network local -- reset
+❌ error: Missing signing key for account GDAQWVA6REGN47BBCFY6SGQ4YTIGMDZZFHDOVUZXMVRAAT6OEZGCACGH
+```
+
+The account called `me` is the admin, Bob is just a regular user. `me` can call `reset`, Bob gets an error.
+
+### Bonus Step 3: Go Pro and Save Bytes with `unsafe`
```
tupui: I am really not sure about showing that. Why not just using `unwrap`? `unwrap_unchecked` goes a level deeper but at the cost of using `unsafe`. I think people should discover super patterns like that on their own otherwise we should put some big read caveats about not abusing that.

website/docs/tutorial/01-getting-started.md
```diff
@@ -239,41 +239,52 @@ impl GuessTheNumber {
...
-    }
+/// Update the number. Only callable by admin.
+pub fn reset(env: &Env) {
+    Self::require_admin(env);
+    let new_number: u64 = env.prng().gen_range(1..=10);
+    env.storage().instance().set(&THE_NUMBER, &new_number);
+}
 ```
 
 And here is the reset function. Note that we use `require_admin()` here so only you can run this function. It generates a random number between 1 and 10 and uses our key to store it.
```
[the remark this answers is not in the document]
tupui: Disregard, I see the full contract now in the part 2.

website/docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,501 @@
...
+        Self::require_admin(env);
+        env.deployer().update_current_contract_wasm(new_wasm_hash);
+    }
+
+    /// Get current admin
+    pub fn admin(env: &Env) -> Option<Address> {
+        env.storage().instance().get(ADMIN_KEY)
+    }
+
+    /// Set a new admin. Only callable by admin.
```
tupui: The comment is slightly different than the implementation. It can only set the admin once and the initial callee is not authorised actually. I think we should add that in the constructor.

website/docs/tutorial/02-making-improvements.md
```diff
@@ -0,0 +1,501 @@
...
+- A more robust `reset` function that uses our helper
+- Better error handling in the `guess` function
+
+## 🪲 Let's break the app!
+
+To understand the bug in our code, let's trigger it. We'll do this by making a small change in `environments.toml`. On our way to finding the line we need to change, we'll learn more about how `environments.toml` works.
+
+Open up `environments.toml` in your editor. Put it side-by-side with the output from `npm run start`. We'll walk through it bit by bit.
+
+### 1. The Network
```
tupui: We are not touching on `.env`. I am not sure if we should do that here too.

website/docs/tutorial/03-adding-payments.md
```diff
@@ -0,0 +1,216 @@
...
+        guesser,
+        x.balance(self.current_contract_address()),
+      );
+      if tx.is_err {
+        panic!("transfer failed!");
+      }
+  } else {
+    // Before transferring their funds, make sure guesser is actually the one calling this function
+    guesser.require_auth();
+    let tx = xlm_client.transfer(guesser, self.current_contract_address(), 1_000_000_0);
```
tupui: We can actually make guess without paying anything because we check the guess before taking money out. They just need to use an empty account and play until they win the pot.

We can either leave it like that and ask people if they can notice a big security issue or fix it with some explanations here.

(For an easy approach, I think we need to do the transfer out, check, and then get the pot if we are correct.)

website/docs/tutorial/03-adding-payments.md
```diff
@@ -0,0 +1,216 @@
...
+  -- guess \
+  --guesser $(stellar keys address bob) \
+  --a_number 3
+
+# Check the pot after Bob's guess
+stellar contract invoke \
+  --id [CONTRACT_ID] \
+  --source alice \
+  --network local \
+  -- get_prize_pot
```
tupui: Some leftover with get_prize_pot

website/docs/tutorial/04-best-practices.md
```diff
@@ -0,0 +1,381 @@
...
+stellar contract invoke \
+  --id [CONTRACT_ID] \
+  --source alice \
+  --network local \
+  -- emergency_withdraw
+```
+
+## Step 8: Gas Optimization?
+
+_🏗️✨ TODO_
```
tupui: Nice to have section on that, we can show `env.cost_estimate().budget()`

website/docs/tutorial/04-best-practices.md
```diff
@@ -0,0 +1,381 @@
...
+#![no_std]
+use admin_sep::{Administratable, Upgradable};
+use soroban_sdk::{
+    contract, contractimpl, contracttype, symbol_short, token,
+    Address, Env, Symbol
+};
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[contracttype]
+pub enum Error {
```
tupui: I would suggest using higher numbers. I am not sure about the range but SAC etc when they panic they use low numbers. E.g. if transfer fails for a low balance, it will be error 10. I know this because I have been bitten by that! If you have in your contract an error 10, then in the frontend you might parse that with your error 10. This gets super confusing.

There is an EIP for that but AFAIK not yet a SEP to define a range and known error codes.

website/docs/tutorial/04-best-practices.md
```diff
@@ -0,0 +1,381 @@
...
+- `obfuscated_value ^ salt = number` (XOR is reversible)
+Without knowing the salt, the stored value looks random. Note this doesn't provide cryptographic security, but makes casual observation much harder.
+
+
+### Advanced techniques
+
+_🏗️✨ TODO: add explanations_
+
+- commit-reveal pattern
+- off-chain oracle
```
tupui: Oracles can be a nice follow up to discuss cross contract calls.

tupui (commented):
I finished, really nice tutorial 😃 👏 

Hope my suggestions help, feel free to discard of course, I don't want to slow things down as we have hackathon where this can really be used.

PR #176, opened 2025-08-28
feat(registry-cli): add download, create-alias, upgrade commands

willemneal:
This allows users to fetch the bytes of a publish Wasm and to upgrade deployed contracts.

PR #199, opened 2025-09-25
feat: remove mentions of stellar registry install

tupui:
Depends on #176

For now just remove mentions of install. I think in the end making the alias should be done by default after `stellar registry deploy`. Just as it is done when calling `stellar contract deploy`.


tupui:
Will be superseded by #212

PR #200, opened 2025-09-25
fix: build clients if they do not exist

pselle:
In the case where you've removed the folders in your packages/ folder, scaffold stellar will never rebuild the clients if the hash matches. Add a check for this.


pselle:
Merged #201 on the 1.86 current main branch

PR #201, opened 2025-09-25
fix: If the packages do not exist, build the clients

pselle:
Draft branch built without updating rust toolchain

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -1022,7 +1022,6 @@ mod tests {
 
     #[test]
     fn test_get_package_dir() {
-        // create a folder at this temporary path for the test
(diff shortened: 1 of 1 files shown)


crates/stellar-scaffold-cli/src/commands/build/clients.rs
```diff
@@ -651,6 +651,18 @@ export default new Client.Client({{
         Ok(())
     }
 
+    fn get_package_dir(&self, name: &str) -> Result<std::path::PathBuf, Error> {
```
chadoh, line 654: This works and if you merge as-is I won't complain but an `Option` might make more sense. It's not really an _error_ if there's no package directory, it just either exists or it doesn't.

pselle: ahhh, yeah, you're correct there. I may go ahead and merge as-is for the bug fix but that Option is nice.

PR #206, opened 2025-09-30
fix: default to development environment

tupui:
The default environment should be development instead of production. The first experience of Scaffold is most likely going to be to develop something locally. Regardless, as a best practice, people must always be deliberate if they intent to deploy on mainnet as to not break things on their side.


pselle:
I agree with this proposal. When I run scaffold without arguments and it defaults to prod, I find that fairly confusing.

@chadoh @willemneal ?

willemneal (approved):

Issue #211, opened 2025-10-06
Macros for Importing contracts and assets


willemneal:
We already provide a way to import a contract via the name of the wasm. The `import_contract_client` macro looks for `target/stellar/<name_used_in_macro>.wasm`.

However, `import_contract` #5 resolves the contract id to allow creating the client at runtime with just `Env`. This means that the contract id is resolved at compile time. Since the contract id is determined based on the  network, we need to update the build process and the current `import_contract_client` to be network aware.

Lastly there is also `stellar_asset` macro from the loam sdk, which generated  SAC addresses for assets at compile time. This too must be network aware.

https://github.com/loambuild/loam/blob/9c6b74e4173e6d9bb3e697aa86924d431f27c0bd/crates/loam-sdk-macro/src/lib.rs#L115-L128

- [ ] Update build to be network aware and put built wasms under `target/<network>/<name>.wasm`
- [ ] Update `import_contract_client` to be a aware of this
- [ ] Finish #5
- [ ] Create `import_stellar_asset` modeled after the example above

tupui:
@willemneal We can close right?

PR #213, opened 2025-10-07
feat: add network in target wasm path

tupui:
Addresses #211


crates/stellar-scaffold-test/src/common.rs
```diff
@@ -342,7 +342,7 @@ pub fn find_binary(name: &str) -> Option<PathBuf> {
 pub fn find_stellar_wasm_dir() -> Option<PathBuf> {
     let exe_path = env::current_exe().ok()?;
     let project_root = find_project_root(&exe_path)?;
-    Some(project_root.join("target").join("stellar"))
+    Some(project_root.join("target").join("stellar/standalone"))
```
willemneal, line 345: Also why standalone here and local above?

tupui: I pushed some changes with a wider replace.

willemneal (approved):

PR #216, opened 2025-10-13
Feat: Copy .env and Run git init 

zachfedor:
During `stellar scaffold init`, copy the `.env.example` to `.env` so variables are available immediately. Also, check if `git` is installed and, if so, run `git init` and make an initial commit.


pselle:
The docs should be updated to reflect this change!

PR #222, opened 2025-10-17
Add release workflow

tupui:
* Release workflow
* npm instead of bun


tupui:
Thanks, yep I will close once after I manage to validate it works

tupui:
Aaaand it failed 😅 I see what is going on, making a fix!

PR #244, opened 2025-10-21
Update domain for website on main branch

pselle:
This is already done on gh-pages branch because of changing from the GitHub UI -- change on main so we don't overwrite it!


tupui:
Getting this in. I will either way do a PR soon with i18n and update other links.

PR #247, opened 2025-10-22
Add i18n infrastructure

tupui:
* i18n
* Basic support for Spanish to get rolling

### DIFF
--- a/website/i18n/es/docusaurus-theme-classic/footer.json
+++ b/website/i18n/es/docusaurus-theme-classic/footer.json
@@ -1,6 +1,6 @@
 {
   "copyright": {
-    "message": "Copyright © 2025 <a href=\"https://theaha.co\">The Aha Company</a>. Built with <a href=\"https://Stellar.org\">Stellar</a>.",
+    "message": "Copyright © 2025 <a href=\"https://theaha.co\">The Aha Company</a>. Creado con <a href=\"https://Stellar.org\">Stellar</a>.",
     "description": "The footer copyright"
   }
 }


website/src/pages/index.tsx
```diff
@@ -76,7 +82,11 @@ export default function Home(): ReactNode {
   return (
     <Layout
       title={`${siteConfig.title}`}
-      description="Scaffold - The Stellar app lifecycle streamlined">
+      description={translate({
+        id: "homepage.mainMessage",
```
chadoh: I don't see any translations in place for all these ids. Could you add them, even if they're in English for now to make it easier to update them for a 1st-time contributor later?

chadoh (changes requested):
Thanks! I haven't tried it out, but it seems like it's probably almost ready. A couple notes below.

website/src/pages/index.tsx
tupui: oh these are in the `code.json`. Or I missed some? I will double check 👍

website/i18n/es/code.json
```diff
@@ -0,0 +1,365 @@
...
+  },
+  "theme.blog.post.paginator.olderPost": {
+    "message": "Publicación más antigua",
+    "description": "The blog post button label to navigate to the older/next post"
+  },
+  "theme.tags.tagsPageLink": {
+    "message": "Ver Todas las Etiquetas",
+    "description": "The label of the link targeting the tag list page"
+  },
+  "theme.docs.breadcrumbs.navAriaLabel": {
```
chadoh: there's also a bunch of keys in here that don't appear anywhere else in this changeset

chadoh (approved):
Thank you!

tupui:
Actually it was just because in the meantime there was the tutorial being added, all good 😃

PR #260, opened 2025-10-29
Try to fix just test-integration

jsmaxi:


jsmaxi:
Before merging, please run the workflows with these changes

zachfedor (approved):

zachfedor:
Thanks @jsmaxi 🙏

zachfedor:
Yup, actively working on that. OpenZeppelin changed some of their example contracts, so we're upgrading to match and adding a safeguard in case it happens again.

Issue #250, opened 2025-10-25
Vendored OpenSSL issue due to missing Perl


jsmaxi:
Install scaffold from source on Windows:
```
cargo install --locked stellar-scaffold-cli
```

Observe that the build fails if Perl is not installed on the machine:
```
Error configuring OpenSSL build:
Command 'perl' not found. Is perl installed?
Command failed: "perl" "./Configure" "--prefix=C:/.../release/build/openssl-sys/out/openssl-build/install" ...
```

Scaffold project uses a vendored OpenSSL, which still requires a working C toolchain and Perl for OpenSSL’s build scripts.
```
[dependencies]
openssl = { version = "0.10", features = ["vendored"] }
```

Suggestion: add a short install requirements section somewhere in the repo to inform users of these dependencies, for example:

```
Install Requirements

This crate uses OpenSSL with the `vendored` feature enabled.
That means OpenSSL will be built from source automatically.

You’ll need these tools installed:

- Windows: [Strawberry Perl](https://strawberryperl.com/) 
- macOS: Xcode command-line tools (`xcode-select --install`)
- Linux: `sudo apt install build-essential perl pkg-config`

If these are already installed, no action is needed, Cargo will compile OpenSSL automatically.
```

---

After installing Perl on Windows, the issue disappeared and the build completed.

willemneal:
Thanks for reporting! Glad it was easy to fix!

PR #263, opened 2025-10-29
Contribute: Add Windows setup instructions and update CONTRIBUTING.md

jsmaxi:


jsmaxi:
Fixes issue #250 and ends the Mac/Linux supremacy.

willemneal:
Thanks! @jsmaxi I do think it's worth trying to figure out why this project depends on the vendored SSL. the stellar CLI doesn't...

willemneal (approved):

willemneal:
Yeah definitely not a blocker for this. Just want to see about removing the need to install perl.
@jsmaxi If you really want to end the unix only support, you can also see about the CI actions for building the artifacts. This would let binstall work which would also allow us to avoid the issues around perl.  If you update the files and make a small change to a `CHANGELOG.md` file of either CLI it will do a dry run of building the artifacts for the different targets in the file. Currently windows is just commented out!

PR #278, opened 2025-11-03
feat: pin scaffold FE repo to specific version

### LINKED ISSUE #254: Add a `--tutorial` flag to `init` command
Out of the box, Scaffold Stellar should create a fully featured template that's immediately ready to deploy and run with fully featured example contracts. But for the tutorial, it's helpful to have a simplified starting point that can be built up into that fully featured example.

## Proposed Idea

1. Duplicate the `guess-the-number` contract in the frontend repo to `tutorial` directory. The former will be maintained as the final product while the latter will be a minimal, but still viable, contract to be built upon via steps in the tutorial.
2. When someone runs `scaffold init` command, we'll check for a `--tutorial` flag:
    - by default, without the flag, we'll delete the extra `tutorial` directory and init works exactly as it does now
    - but with the flag, we'll remove all contracts _except_ the tutorial directory and rename it to be `guess-the-number` to be a simplified starting point ready to be improved

As for documentation, the README, the home page, and the quickstart don't need to mention this flag. It only needs to be used on the tutorial when introducing `init` with a comment that it provides different output than normally.

zachfedor:
Upgrade to new version of degit that accepts tags and use that for more deterministic `scaffold init` results. Also shares the constant with the upgrade command so it's only defined once.

This will allow us create separate branches/tags in the frontend repo for things like the tutorial or for easily testing changes across both projects.

Usage:
```sh
$ stellar scaffold init my-project  # uses main branch of scaffold-stellar-frontend
$ stellar scaffold init --tutorial my-project  # use tutorial branch
$ stellar scaffold init --tag v1.0.0 my-project  # use specified tag

$ stellar scaffold init --tag v1.0.0 --tutorial my-project  # fails, can only specify tag or tutorial, not both.
```

Fixes #254

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -941,7 +941,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


pselle:
@zachfedor Understand the want to point to a tutorial branch, 👍 for that part.

In the future, we might want to point to "download the latest tagged frontend" and perhaps might even implement that when we allow people to BYO frontend… but for now "whatever is on main" is working well for release speed.

So I think, "yes" to your question, but that the code would need to update to reflect that.

PR #279, opened 2025-11-03
fix(contracts/registry): Clean up based on Audit feedback

willemneal:
- Make some contract methods pub(crate)
- Also improve Key trait name
- get -> get_contract_id for clarity
- Refactor name canonicalization to improve clarity and ensure that keyword check happens after normalization
- Ensure that the hash keys have a unique type prefix so there is no chance of key namespace overlap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,44 +4,38 @@ version = 3
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "admin-sep"
 version = "0.0.0"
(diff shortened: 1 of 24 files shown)


contracts/registry/src/name/normalized.rs
```diff
@@ -0,0 +1,169 @@
...
+pub(crate) struct Normalized {
+    len: usize,
+    internal: [u8; Self::MAX_NAME_LENGTH],
+}
+
+impl Normalized {
+    pub const MAX_NAME_LENGTH: usize = 64;
+
+    pub fn canonicalize(s: &String) -> Result<String, Error> {
+        let normalaized_str = Normalized::new(s)?;
```
pselle:
```suggestion
        let normalized_str = Normalized::new(s)?;
```
You have a typo in the variable name

contracts/registry/src/test.rs
```diff
@@ -245,39 +245,61 @@ fn returns_most_recent_version() {
...
-        if result {
-            assert!(raw_result.is_ok(), "should be valid: {s}");
-        } else {
-            assert_eq!(
-                raw_result,
-                Err(Error::InvalidName),
-                "should be invalid: {s}"
-            );
-        }
+fn test_string(s: &str, result: bool) {
```
pselle: This doesn't have a `#[test]` annotation. Should it?

pselle (changes requested):
A few typos and a couple Qs about "did you mean to remove this test macro annotation"

PR #283, opened 2025-11-04
fix: remove unneeded stellar-cli features

willemneal:
Currently this pulls in wasm-opt and ledger and other features. This would be better to depend on stellar-cli's binary rather than as a libraries since it bloats our binary and makes builds more complex and take longer.

Also adds back all targets for binaries.


willemneal:
Once #282 is merged. This one should be too and then we will have all targets building and like 300 fewer dependencies!

PR #315, opened 2025-11-22
Asset macro rework

ifropc:
This PR reworks asset marco to match POC from @willemneal work done [here](https://github.com/theahaco/scaffold-stellar-frontend/blob/main/contracts/guess-the-number/src/xlm.rs)

It introduces 2 mods, 1 for unit testing, the other for actual "production" code (basically any env that is not a mock one. For example, can be used in contracts)

The unit testing mod is largely the same as the snapshot above (simply deploys new SAC and uses its instance to get clients). Useful to mock actual SACs in tests. (e.g. for XLM or any other asset).

The production mod deploys an actual SAC contract (if needed) and gets SAC contract ID using generated serialized asset (from the asset code). There are limitations on "production" mod for unit tests, so it's largely unusable, unfortunately (can't establish trustlines, can't mint)

### DIFF
--- a/crates/stellar-scaffold-macro/src/asset.rs
+++ b/crates/stellar-scaffold-macro/src/asset.rs
@@ -108,7 +108,7 @@ pub fn parse_literal(lit_str: &syn::LitStr) -> TokenStream {
                 soroban_sdk::token::TokenClient::new(&env, &contract_id(env))
             }
 
-            pub fn register(env: &soroban_sdk::Env, admin: &soroban_sdk::Address) {
+            pub fn register(env: &soroban_sdk::Env) {
                 let symbol = token_client(env).try_symbol();
                 if symbol.is_err()  {
                     env.deployer().with_stellar_asset(SERIALIZED_ASSET).deploy();
@@ -214,7 +214,7 @@ mod test {
                     soroban_sdk::token::TokenClient::new(&env, &contract_id(env))
                 }
 
-                pub fn register(env: &soroban_sdk::Env, admin: &soroban_sdk::Address) {
+                pub fn register(env: &soroban_sdk::Env) {
                     let symbol = token_client(env).try_symbol();
                     if symbol.is_err()  {
                         env.deployer().with_stellar_asset(SERIALIZED_ASSET).deploy();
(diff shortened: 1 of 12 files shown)


crates/stellar-scaffold-macro/src/asset.rs
```diff
@@ -31,38 +37,68 @@ pub fn parse_asset(str: &str) -> Result<(xdr::Asset, String), xdr::Error> {
...
+            /// Registers a new SAC contract (to use in unit tests only)
+            pub fn register(env: &soroban_sdk::Env, admin: &soroban_sdk::Address) -> soroban_sdk::testutils::StellarAssetContract {
+                let sac = env.register_stellar_asset_contract_v2(admin.clone());
+                let cl = stellar_asset_client(env, &sac);
+                env.mock_all_auths();
+                cl.mint(admin, &1_000_000_000_i128);
+                sac
+            }
+
+            pub fn to_min_unit(float: f64) -> i128 {
```
willemneal: Floats are not allowed by the soroban VM. So would be good to have a test with a compiled wasm too.

crates/stellar-scaffold-macro/src/asset.rs
```diff
@@ -97,75 +141,70 @@ mod test {
...
+                    soroban_sdk::token::TokenClient::new(&env, &sac.address())
+                }
+
+                /// Registers a new SAC contract (to use in unit tests only)
+                pub fn register(env: &soroban_sdk::Env, admin: &soroban_sdk::Address) -> soroban_sdk::testutils::StellarAssetContract {
+                    let sac = env.register_stellar_asset_contract_v2(admin.clone());
+                    let cl = stellar_asset_client(env, &sac);
+                    env.mock_all_auths();
+                    cl.mint(admin, &1_000_000_000_i128);
+                    sac
```
willemneal: We should also reset the auths by passing an empty array to set auths. This way the caller won't have this side effect.

willemneal (changes requested):

crates/stellar-scaffold-macro/src/asset.rs
```diff
@@ -31,38 +37,68 @@ pub fn parse_asset(str: &str) -> Result<(xdr::Asset, String), xdr::Error> {
...
 /// Generate the code to read the `STELLAR_NETWORK` environment variable
 /// and call the `generate_asset_id` function
-pub fn parse_literal(lit_str: &syn::LitStr, network: &Network) -> TokenStream {
-    let (contract_id, code) = generate_asset_id(&lit_str.value(), network).unwrap();
-    // let contract_id = format_ident!("\"{contract_id}\"");
-    let contract_id = contract_id.to_string();
+pub fn parse_literal(lit_str: &syn::LitStr) -> TokenStream {
+    let (code, data) = get_serialized_asset(&lit_str.value()).unwrap();
     let mod_name = format_ident!("{code}");
+    let test_mod_name = format_ident!("test_{code}");
```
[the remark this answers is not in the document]
ifropc: Maybe I'm wrong but isn't cfg test not being exported from the binary? 
We could do a opt-in testutils like soroban sdk does

PR #322, opened 2025-12-04
feat: add managed contracts and `stellar-registry-build`

willemneal:
# Changes to Registry Contract

## Managed registries

The biggest change is now when constructing the contract  there is an argument for an optional manager. If provided (or added later), the manager account must authorize any initial publishes or deploying or registering contract names. This way we can have an initial root registry which only contains verified published wasms and named contract, which allows us and ecosystem partners to ensure that current established contracts will be in the verified registry.

If the manager is not provided then anyone can publish or register a contract name with a registry.

## Root registry

The root registry is managed and in the constructor if `is_root` is true, deploys the `unverified` registry as unmanaged and registers the name.

## Name resolution

Since there are now multiple registries the CLI and other tooling will resolve names using the root registry as the source of truth. For example, `unverified/hello_world` would first fetch the contract id for `unverified` from the root registry and then search in that registry for `hello_world`.

## `NormalizedName`

This new type helps to ensure that the names for publishing and registering contract ids are valid and normalized, e.g. `contract_name` becomes `contract-name` . It is used as keys in storage, this way inputs are forced to be parsed. As @ifropc had shown in his recent PR it is easy to miss a path if not done this way.

## `register_contract` and `fetch_contract_owner`

The other big change to the contract is a method to register a contract name without deploying the contract through the registry. This is required to support existing contracts and those who want to handle deploying themselves. To support this the storage was updated to track the owner account which registered the name. And a new `Register` event was added. This means that when deploying through the contract a deploy and register event is made.

`fetch_contract_owner` was also added to allow looking up the owner who registered the contract name.

## `deploy_unnamed`

Since the `root` contract is now managed, we added a way to deploy a published wasm without registering a contract name. This allows a few possibilities:

- A project that wants to manage Wasm binaries on Registry while obfuscating their contract itself. Perhaps they are not ready to have a public-facing contract yet, or perhaps they never will want their contract to be public.
- People, especially learners, who want to deploy a throwaway contract directly from the Wasm Registry, without having a permanent entry show up for this contract in the Contract Registry.

# Changes to `stellar-registry-cli`

## Using new name resolution

The contract and wasm name resolution was added to the registry cli commands anywhere a name is used. So for example,

```bash
stellar registry fetch_wash_hash --wasm-name unverified/hello_world
```

— this will fetch a contract called `hello-world` from the `unverified` registry.

## New commands
The following commands were added to support the available contract methods:
- `deploy-unnamed` 
- `fetch-contract-id`
- `fetch-contract-owner`
- `publish-hash` 
- `registry-contract`

# New `stellar-registry-build` crate

This crate helps to resolve contracts both at build time and for the `stellar-registry-cli`. As mentioned above, resolving a contract name may now involve searching multiple registries, e.g. `unverified/hello_world`.  It can also resolves contracts the same way as stellar CLI, via the deployer and salt. 

For example:
```rust
import_from_registry("unverified/hello_world");
import_from_alias("hello_world");
import_from_deployer("hello_world:G233...");
```

# Add Github Attestation 
The new github action and workflow which is triggered by any tag with the pattern `registry-v*` and publishes a release including a wasm file with an attestation. A `from-github` option was added to the registry CLI`s `publish` command, which downloads the wasm file with the added github attestation to publish.

### DIFF
--- a/contracts/registry/src/registry/contract.rs
+++ b/contracts/registry/src/registry/contract.rs
@@ -136,7 +136,7 @@ impl Contract {
     /// Since we know that this contract is an executable we can skip checking when unwrapping
     /// which is unsafe.
     ///
-    /// Furthermore it uses the NormalizedName::new_unchecked, which is unsafe because it skips validating
+    /// Furthermore it uses the `NormalizedName::new_unchecked`, which is unsafe because it skips validating
     /// the name, which we know already to be valid.
     pub(crate) fn deploy_unverified_and_claim_registry(env: &Env, admin: &Address) {
         unsafe {


website/docs/deploy.md
```diff
@@ -2,32 +2,42 @@
 
 ## Smart contract
 
-When you are ready for testnet/mainnet, we recommend to deploy your contract using `stellar registry`. Some commands to get you started.
+When you are ready for testnet/mainnet, we recommend to deploy your contract using `stellar registry`. The registry has two namespaces:
+
+- **Verified registry** (default) - Requires manager approval for initial publishes
```
willemneal:
```suggestion
- **Verified registry** (default) - Requires manager approval for initial publishes and deploys
```

website/docs/quick-start.mdx
```diff
@@ -51,26 +51,28 @@ You should see your React frontend at http://localhost:5173.
...
 # Can access the help docs for constructor parameters
 stellar registry deploy \
-  --contract-name my-contract-instance \
-  --wasm-name my-contract \
+  --contract-name unverified/my-contract-instance \
+  --wasm-name unverified/my-contract \
   -- \
   --help
 
 # Install the deployed contract locally for use with stellar-cli
```
willemneal:
```suggestion
# Create an alias the deployed contract locally for use with stellar-cli
```

website/docs/quick-start.mdx
```diff
@@ -112,27 +114,33 @@ stellar scaffold build
 **stellar-registry** Manage contract deployment and versions:
 
 ```
-stellar registry publish --wasm contract.wasm --wasm-name my-contract    # Publish contract to the registry
-stellar registry deploy --contract-name instance --wasm-name my-contract # Deploy a contract instance
-stellar registry create-alias my-contract-instance                       # Install deployed contracts locally
+# Publish to unverified registry (no approval needed)
+stellar registry publish --wasm contract.wasm --wasm-name unverified/my-contract --binver "1.0.0"
```
willemneal:
```suggestion
stellar registry publish --wasm contract.wasm --wasm-name unverified/my-contract
```

crates/stellar-build/src/networks.rs
```diff
@@ -18,7 +18,7 @@ pub enum Error {
         r#"Invalid STELLAR_PASSPHRASE: {0}. 
             Expected: "Standalone Network ; February 2017",
                       "Test SDF Network ; September 2015",
-                      "Future SDF Network ; September 2023",
+                      "Test SDF Future Network ; October 2022",
```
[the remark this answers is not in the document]
willemneal: This is actually a bug. According to the official passphrases this change is correct.

crates/stellar-registry-cli/src/commands/deploy.rs
```diff
@@ -7,31 +7,41 @@ pub use soroban_spec_tools::contract as contract_spec;
...
 #[derive(Parser, Debug, Clone)]
 pub struct Cmd {
-    /// Name of contract to be deployed
+    /// Name of contract to be deployed. Can use prefix of not using verified registry.
+    /// E.g. `unverified/<name>`
     #[arg(long, visible_alias = "deploy-as")]
-    pub contract_name: String,
-    /// Name of published contract to deploy from
+    pub contract_name: PrefixedName,
+    /// Name of published contract to deploy from. Can use prefix of not using verified registry.
```
elizabethengelman:
```suggestion
    /// Name of published contract to deploy from. Can use prefix if not using verified registry.
```

elizabethengelman (commented):
👏 🎉 Haven't tested this out locally yet, but it looks awesome!

I left a few comments & questions, but I don't think any of them are merge blocking.

contracts/registry/src/registry/contract.rs
```diff
@@ -58,89 +83,200 @@ impl Contract {
         let _ = r.map_err(|_| Error::UpgradeInvokeFailed)?;
         Ok(contract_id)
     }
-}
 
-#[contractimpl]
-impl Deployable for Contract {
-    fn deploy(
+    pub(crate) fn register_contract_name(
```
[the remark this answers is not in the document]
willemneal: This isn't part of the public facing API so depending on whether it is verified or not, the manager will need to approve before this method is called. E.g.
```rust
Contract::assert_no_contract_entry_and_authorize(env, &admin, &contract_name)?;
```
Otherwise it is used in the constructor.

contracts/registry/src/registry/contract.rs
```diff
@@ -58,89 +83,200 @@ impl Contract {
...
         .publish(env);
-
         Ok(contract_id)
     }
 
-    fn fetch_contract_id(env: &Env, contract_name: String) -> Result<Address, Error> {
-        let contract_name = canonicalize(&contract_name)?;
-        Self::get_contract_id(env, &contract_name)
+    pub(crate) fn deploy_unverified_and_claim_registry(env: &Env, admin: &Address) {
+        unsafe {
```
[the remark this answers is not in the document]
willemneal: Very true. I'll add a comment!

PR #352, opened 2026-01-16
Feat/scaffold clean

### LINKED ISSUE #259: `stellar scaffold reset` to clear all artifacts
Add `stellar scaffold reset` command to clear all scaffold-generated artifacts from:

- `target/stellar`
- contract aliases (used to be able to just `rm .stellar` directory, but with only-global config this no longer works; we will need to find a different way)
- `packages/*`  (but not checked-into-git files like `.gitkeep`)
- `src/contracts/*` (but not checked-into-git files like `util.ts`)

elizabethengelman:
closes #259 

Thinking of using this instead of the copilot created PR https://github.com/theahaco/scaffold-stellar/pull/302

This PR pulls some stuff from that PR, and does a few things differently.

### DIFF
--- a/.github/workflows/cd-registry-cli.yml
+++ b/.github/workflows/cd-registry-cli.yml
@@ -57,10 +57,10 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         with:
           path: |
             ~/.cargo/bin/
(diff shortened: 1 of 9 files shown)


crates/stellar-scaffold-cli/src/commands/clean.rs
```diff
@@ -0,0 +1,422 @@
...
+        Self::clean_dir(
+            workspace_root,
+            &src_contracts_path,
+            git_tracked_src_contract_entries,
+            printer,
+        )
+    }
+
+    fn clean_contract_aliases(workspace_root: &Path, printer: &Print) -> Result<(), Error> {
+        match Environment::get(workspace_root, &ScaffoldEnv::Development) {
```
elizabethengelman: Instead of checking if .env is setting XDG_CONFIG_HOME or not, I think that we can skip that just go straight to checking the environment.toml file for contracts that have been created, and then use `stellar contract alias remove` to delete them.

PR #376, opened 2026-02-06
Feat/guarded registry 5

elizabethengelman:


chadoh:
Thanks for breaking this PR into smaller ones, @elizabethengelman. After talking it over with @pselle, I've been convinced that we can merge the original and verify that everything works correctly afterwards, rather than understanding and verifying prior to merge.

PR #378, opened 2026-02-07
fix: Speed up tests

willemneal:
# Speed up testing

This PR first builds everything and then splits the RPC tests into groups to parallelize them. This also helps speed up subsequent builds because it caches the build and uses sscache to cache built dependencies.

Also it uses reflink (copy-on-write) clones instead of full copies when setting
up test fixtures, making test setup near-instant on APFS/btrfs with
automatic fallback to normal copies on unsupported filesystems.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


willemneal:
@chadoh It seems that there the example tests are flaky. My guess is that it has something to do with the github outage since it seems to be an issue with caching the repo.

PR #386, opened 2026-02-11
fix: reorganize RPC tests to simplify filters and fix flaky example tests

willemneal:
- Now filters are folders or files; making it clearer what is being run.
- Run a single example test first to warm up a shared cache
- use `--locked` to improve cache reuse since builds will be more likely to share deps with main if `Cargo.lock` hasn't been updated.

### DIFF
--- a/justfile
+++ b/justfile
@@ -42,7 +42,6 @@ build-cli-test-contracts:
     just stellar-scaffold build --manifest-path crates/stellar-scaffold-test/fixtures/soroban-init-boilerplate/Cargo.toml
 
 test: build
-    cargo t -E 'package(stellar-scaffold-cli)'
     cargo t
 
 test-integration: build-cli-test-contracts


justfile
```diff
@@ -41,39 +42,45 @@ build-cli-test-contracts:
...
+test-integration-scaffold-examples-1 ci="false":
+    just _test-scaffold 'test(examples::) and test(/case_01/)' {{ ci }}
+    just _test-scaffold 'test(examples::) and (test(/case_0[2-9]/) or test(/case_1[0-4]/))' {{ ci }}
 
 # Run scaffold-cli example integration tests (cases 15-27)
-test-integration-scaffold-examples-2:
-    just _test-scaffold 'test(examples::) and (test(/case_1[5-9]/) or test(/case_2/))'
+test-integration-scaffold-examples-2 ci="false":
+    just _test-scaffold 'test(examples::) and test(/case_15/)' {{ ci }}
+    just _test-scaffold 'test(examples::) and (test(/case_1[6-9]/) or test(/case_2/))' {{ ci }}
```
chadoh: Same question here

```suggestion
    just _test-scaffold 'test(examples::) and (test(/case_1[5-9]/) or test(/case_2/))' {{ ci }}
```

PR #392, opened 2026-02-19
fix: update docker container running alert

zachfedor:
Match an update to the Stellar-CLI from [this PR](https://github.com/stellar/stellar-cli/pull/2234) that broke how we determined if a container was already running. Currently, `scaffold build` will error out if the container is already running because the string mismatch, or logic assumes there was an error with it instead of continuing on with the health check.

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/build/clients.rs
+++ b/crates/stellar-scaffold-cli/src/commands/build/clients.rs
@@ -756,7 +756,7 @@ export default new Client.Client({{
         let deploy_cmd = cli::contract::deploy::wasm::Cmd::parse_arg_vec(&deploy_arg_refs)?;
         let contract_id = deploy_cmd
             .execute(
-                &deploy_cmd.config,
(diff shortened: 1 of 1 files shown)


willemneal (approved):

zachfedor:
@willemneal, it took me a while to hunt this down, but I think I found a problem from your PR that updated scaffold to use stellar-cli v25, [specifically these lines](https://github.com/theahaco/scaffold-stellar/commit/e2f65257299b94bd7ef74d2f842215fc20594b34#diff-25348984e574ea5febb29a3894695b7a58fd32b7f3d26f7515a8264f580f3e99R756-R762). I think the `deploy_contract` function was using the wrong network configuration, and since that didn't include `--rpc-url` or `--network-passphrase`, the parsed config had no network info which was causing `get_remote_wasm_from_hash` to fail with "Contract Code not found" on my system when trying to build clients.

No need to look now, I can explain better tomorrow in our weekly update.

PR #406, opened 2026-03-06
feat(macro): add registry fallback for import_contract_client

willemneal:
## Summary

- `import_contract_client!` now falls back to on-chain registry when local wasm is not found
- Resolution order: local build target → `stellar registry download` → panic with clear message
- Set `STELLAR_NO_REGISTRY=1` to disable registry fallback for offline builds

## Test plan

- [x] Existing macro tests pass
- [x] `cargo clippy` clean
- [x] Publish wasm to local registry, remove local build, verify `import_contract_client!` fetches from registry
- [x] Verify `STELLAR_NO_REGISTRY=1` disables fallback

### DIFF
--- a/crates/stellar-scaffold-macro/src/lib.rs
+++ b/crates/stellar-scaffold-macro/src/lib.rs
@@ -1,9 +1,12 @@
 #![recursion_limit = "128"]
 extern crate proc_macro;
 use proc_macro::TokenStream;
+use proc_macro2::Ident;
 use quote::quote;
 use std::env;
 use stellar_build::Network;
+use syn::parse::{Parse, ParseStream, Result};
+use syn::parse_macro_input;
 
 mod asset;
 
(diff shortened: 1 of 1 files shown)


crates/stellar-scaffold-macro/src/lib.rs
```diff
@@ -46,6 +41,50 @@ pub fn import_contract_client(tokens: TokenStream) -> TokenStream {
...
+}
+
+fn download_from_registry(
+    name: &str,
+    target_dir: &std::path::Path,
+) -> Result<std::path::PathBuf, Box<dyn std::error::Error>> {
+    let output = target_dir.join(name).with_extension("wasm");
+    if let Some(parent) = output.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
```
willemneal: If it didn't have a parent we would be in the root directory, which seems like an error to me.

PR #442, opened 2026-03-30
feat(macro): accept string literals in import_contract_client

### LINKED ISSUE #440: can't use unverified registry with `stellar_registry`  macro?
This works great:

```rs
stellar_registry::import_contract_client!(registry);
```

But if I try to use an unverified Wasm listed at https://testnet.rgstry.xyz/wasms, such as:

```rs
stellar_registry::import_contract_client!(unverified/guess-the-number);
```

When I save the file, Rust autofmt renders it:

```rs
stellar_registry::import_contract_client!(unverified / guess - the - number);
```

...which of course gives a "not found" error. If I try to fix it by using a string:

Image

There also seems to be no way to specify `unverified` using some alternate syntax (which, to be clear, I don't think is a _good_ interface, but I thought _might_ be an _available_ interface):

```rs
// does not work! do not want!
stellar_registry::import_contract_client!(guess_the_number, channel: "unverified");
```

Can we make the macro accept a TokenStream _or_ a &str?

willemneal:
## Summary

Fixes #440

- The `import_contract_client!` macro now accepts string literals in addition to bare identifiers
- Enables channel-prefixed registry paths: `import_contract_client!("unverified/guess-the-number")`
- Enables hyphenated contract names: `import_contract_client!("guess-the-number")`
- Backward compatible: `import_contract_client!(registry)` continues to work

When using a string literal, the module name is derived from the contract name with hyphens replaced by underscores (e.g., `guess_the_number`). The full name including channel prefix is passed to `stellar registry download` for correct resolution.

## Test plan

- [ ] Verify `import_contract_client!(registry)` still works (backward compat)
- [ ] Verify `import_contract_client!("unverified/guess-the-number")` resolves correctly via registry
- [ ] Verify `import_contract_client!("guess-the-number")` works for hyphenated names
- [ ] Verify rustfmt does not corrupt string literal input
- [ ] `cargo clippy -p stellar-scaffold-macro` passes clean

### DIFF
--- a/crates/stellar-scaffold-macro/src/lib.rs
+++ b/crates/stellar-scaffold-macro/src/lib.rs
@@ -57,6 +57,83 @@ struct WasmBinary {
     pub file: String,
 }
 
+#[derive(Debug)]
+#[cfg_attr(not(test), allow(dead_code))]
+struct ParsedInput {
+    /// Optional channel prefix (e.g., "unverified")
+    channel: Option<String>,
+    /// Contract name without channel prefix (e.g., "guess-the-number")
+    contract_name: String,
+    /// Sanitized name suitable for use as a Rust module identifier
+    module_name: String,
+    /// Full input string including channel prefix (e.g., "unverified/guess-the-number")
+    full_name: String,
+}
+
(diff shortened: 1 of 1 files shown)


chadoh:
Closing in favor of #443, which now implements the same set of tests

PR #463, opened 2026-04-10
Feat/reporter extension

### LINKED ISSUE #328: Clean up `--build-clients` output
We should have three kinds of output options:

- Default: optimized output, useful info for most devs in most situations
- Verbose: everything, useful for debugging
- Quiet: should be ACTUALLY quiet

### LINKED ISSUE #449: `reporter` extension
Clean up the logic for our contract explorer dashboard, now that we can lean on the extension system.

A good proof-of-concept to demonstrate the power of the extension system, while allowing us to dogfood our own infrastructure.

Will clean up `--build-clients` output (#328) and implement a standalone contract explorer process (#368)

zachfedor:
Adds the reference extension `stellar-scaffold-reporter` using all the new hooks for Scaffold:
- It reports compile timing, Wasm sizes with deltas, deploy details, codegen timing, a summary after each life-cycle phase
- This cleans up the current build output so nothing is duplicated (fixes #328). Now scaffold logging is only essentials (i.e. successes, warnings, errors).
- Adds configured modes "standard" and "minimal" to quiet the output even more
- Adds extension system tests using the reporter

Closes #449

### DIFF
--- a/crates/stellar-scaffold-cli/src/extension.rs
+++ b/crates/stellar-scaffold-cli/src/extension.rs
@@ -34,9 +34,10 @@ pub fn discover(entries: &[ExtensionEntry], printer: &Print) -> Vec<ResolvedExte
 /// Runs a single lifecycle hook across all registered extensions.
 ///
 /// For each extension whose manifest lists `hook`, spawns
-/// `stellar-scaffold-<name> <hook>` as a subprocess, serializes `context` as
-/// JSON to its stdin, waits for it to exit, then forwards its stdout to
-/// Scaffold's own stdout.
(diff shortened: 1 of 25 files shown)


crates/stellar-scaffold-cli/src/extension.rs
```diff
@@ -65,6 +67,17 @@ pub async fn run_hook<C: serde::Serialize>(
             continue;
         }
```
willemneal, line 68: Missed in my last review, but you can do `extensions.iter().(filter)`. Also I would advocate for `all not` vs `not any`, to avoid the `!` in the front.

willemneal (approved):
LGTM! Very nice! Makes things cleaner and is fun!

PR #471, opened 2026-04-14
refactor(registry): use soroban-sdk-tools::scerr for error enum

willemneal:
Adopt soroban-sdk-tools 0.1.2 and replace the manual #[contracterror]
enum with #[soroban_sdk_tools::scerr], letting the macro assign
discriminants automatically.

Agent-Id: bitswell
Session-Id: 8b3f349c-a88e-4d1d-8ef4-df90d110111d

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


contracts/registry/src/registry/contract.rs
```diff
@@ -197,6 +197,11 @@ impl Contract {
                     contract_id: contract_address.clone(),
                 }
                 .publish(env);
+                events::SubRegistry {
```
ifropc, line 200: Seems like a duplicate

contracts/registry/src/error.rs
```diff
@@ -1,48 +1,44 @@
-use soroban_sdk::{self, contracterror};
-
-#[contracterror]
-#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
-#[repr(u32)]
+#[soroban_sdk_tools::scerr]
```
ifropc, line 1: I assume this is backward compatible and works pretty much like go "enums"?
Also, does scerr start from 0 or from 1?

contracts/registry/src/registry/contract.rs
willemneal: I am failing at using the new github stacked branches and this ended up in the wrong PR! It is for registering the registry itself!

contracts/registry/src/error.rs
willemneal: Yep!

chadoh (approved):

PR #478, opened 2026-04-16
feat(registry): update to new contract id for registry

willemneal:
## Summary

Bumps the registry contract to `v0.6.1` and switches the verified contract ID derivation to a `.salt` file so future bumps don't require code changes. Also overhauls the deploy scripts (dry-run, idempotent re-deploy, shellcheck) and fixes a constructor event ordering bug.

## Registry contract

- Bump `registry` crate `0.6.0` → `0.6.1` -> `0.6.2` (updated to the soroban-sdk v26).
- **Constructor event order fix** (`contracts/registry/src/registry/contract.rs`): publish the root `SubRegistry` event before the sub-registry's event and before name registration, so the root event appears first in the event stream.

## Verified contract ID is now driven by `.salt`

- New `contracts/registry/.salt` file (`2026-04-24T13:41:00`) is the single source of truth for the salt used to derive the verified registry contract address.
- `verified_contract_id` (`crates/stellar-registry-build/src/registry.rs`) now `include_str!`s `.salt` instead of the hardcoded `v0.4.1` string.
- Updated expected contract IDs across futurenet / testnet / mainnet / local.

## Test refactor: `expect-test`

- Migrated `generate_id` tests to [`expect-test`](https://crates.io/crates/expect-test) so the four passphrase IDs can be regenerated in-place after a salt bump via `UPDATE_EXPECT=1`.
- Added `expect-test = "1.5"` as a dev-dep of `stellar-registry-build`.
- New `just update-registry-tests` recipe wraps the regeneration command.

## Deploy script overhaul

- Deleted top-level `deploy_registry.sh` (used hardcoded `v0.4.1`, no dry-run).
- New `contracts/registry/deploy.sh`:
  - `--dry-run` / `-n` flag prints commands instead of executing.
  - Reads version from `Cargo.toml`, downloads matching wasm from the GitHub release, computes salt from `.salt` via `shasum -a 256`.
  - Deploys the registry alias, then publishes the wasm.
- `contracts/registry/deploy_initial_registries.sh`:
  - Added `--dry-run` support.
  - Builds and uses the local `stellar-registry-cli` so the deploy uses the same client as this branch.
  - Skips deploy when the named contract is already registered (idempotent re-runs).
  - Passes the root registry id via `--root`.
  - Calls `process_batch --limit 10` after `batch-register`.
  - Skips `batch_register` cleanly when a project has zero initial contracts.
- `initial_contracts.json`: added empty `oz` project entry.

## Shellcheck

- New `check-shell.sh` at repo root: discovers all `*.sh` (excluding `target` / `node_modules`) and runs `shellcheck -x` on each.
- Resolved warnings flagged by shellcheck:
  - `deploy.sh`: dropped a useless `cat` (SC2002).
  - `deploy_initial_registries.sh`: removed dead `INITIAL_BATCH` variable (SC2034).

## Files changed

- `Cargo.lock`, `crates/stellar-registry-build/Cargo.toml` — `expect-test` dep
- `crates/stellar-registry-build/src/registry.rs` — `.salt`-based contract id + test refactor
- `contracts/registry/{Cargo.toml,.salt}` — version bump + salt source
- `contracts/registry/src/registry/contract.rs` — constructor event order fix
- `contracts/registry/deploy.sh` (new), `deploy_initial_registries.sh`, `initial_contracts.json`
- `deploy_registry.sh` (deleted)
- `check-shell.sh` (new), `justfile` — shellcheck + new recipe

## Test plan

- [ ] `just test` passes
- [ ] `UPDATE_EXPECT=1 cargo test --package stellar-registry-build registry::generate_id` is a no-op (expected IDs already current)
- [ ] `./check-shell.sh` passes
- [ ] `contracts/registry/deploy.sh --dry-run` prints the expected `stellar contract deploy` + `just registry publish` invocations
- [ ] `contracts/registry/deploy_initial_registries.sh --dry-run` prints expected per-project deploy + batch-register + process_batch invocations and skips already-registered contracts


chadoh (approved):

contracts/registry/deploy.sh
```diff
@@ -0,0 +1,73 @@
...
+
+VERIFED=$(shasum -a 256 < "$SCRIPT_DIR/.salt" | awk '{print $1}')
+
+ADMIN=theahaco
+ADDRESS=GAMPJROHOAW662FINQ4XQOY2ULX5IEGYXCI4SMZYE75EHQBR6PSTJG3M
+
+run stellar contract deploy --alias registry \
+                        --wasm "$WASM_PATH" \
+                        --source "$ADMIN" \
+                        --salt "$VERIFED" \
```
chadoh: Kinda weird that the variable name is still `VERIFIED`

```suggestion
SALT=$(shasum -a 256 < "$SCRIPT_DIR/.salt" | awk '{print $1}')

ADMIN=theahaco
ADDRESS=GAMPJROHOAW662FINQ4XQOY2ULX5IEGYXCI4SMZYE75EHQBR6PSTJG3M

run stellar contract deploy --alias registry \
                        --wasm "$WASM_PATH" \
                        --source "$ADMIN" \
                        --salt "$SALT" \
```

contracts/registry/deploy.sh
```diff
@@ -0,0 +1,73 @@
...
+                        -- \
+                        --admin "$ADMIN" \
+                        --manager "\"$ADDRESS\""
+
+echo "sleeping so dynamic table can populate on goldsky"
+sleep 60
```
chadoh: Don't need this anymore now that we do the check+refresh step on the indexer side

```suggestion
```

contracts/registry/deploy_initial_registries.sh
```diff
@@ -1,37 +1,77 @@
...
 deploy () {
     local name="$1"
-    stellar registry deploy --contract-name "$name" --wasm-name registry -- --admin theahaco --manager "$ADMIN"
-    stellar registry create-alias "$name"
+    if existing_id=$(stellar registry fetch-contract-id "$name" 2>/dev/null) && [ -n "$existing_id" ]; then
+        echo "Contract '$name' already registered (id: $existing_id); skipping deploy"
+    else
+        run stellar registry deploy --contract-name "$name" --wasm-name registry -- \
+                                    --admin $ADMIN \
+                                    --manager "$MANAGER" \
```
chadoh, line 52: Why can `--admin` be `theahaco` but `--manager` needs to be the pubkey?

justfile
```diff
@@ -100,3 +100,7 @@ clippy *args:
 
 clippy-test:
     just clippy --tests --all-features
+
+# tests which check the determinstic contract names need to be rebuild after changes
```
chadoh: Not totally sure what you meant here. Maybe:

```suggestion
# update deterministic contract IDs after changes to `contracts/registry/.salt`
```

crates/stellar-registry-build/src/registry.rs
```diff
@@ -77,49 +77,52 @@ pub fn contract_id(network_passphrase: &str, salt: &str) -> stellar_strkey::Cont
...
     use stellar_cli::config::network::passphrase::*;
 
-    fn test_contract_id((passphrase, contract_id): (&str, &str)) {
-        assert_eq!(
-            &super::verified_contract_id(passphrase).to_string(),
-            contract_id,
-            "{passphrase}"
-        );
+    /// Run with `UPDATE_EXPECT=1 cargo test` to regenerate the expected contract
+    /// IDs in-place after bumping the registry version.
```
chadoh:
```suggestion
    /// IDs in-place after bumping the registry version or src/registry/.salt
```

contracts/registry/initial_contracts.json
```diff
@@ -1,4 +1,5 @@
 [
+    { "oz": [] },
```
chadoh, line 2: Do you have a separate script where you create the initial wasms in the `oz` registry? Will that be part of a separate PR, or should it be included here?

chadoh (changes requested):
Some quick updates & questions before we merge.

contracts/registry/deploy_initial_registries.sh
willemneal: This is a bug in the CLI. It can handle looking up addresses but  not when they are in an option. I will make an issue for it.

contracts/registry/initial_contracts.json
willemneal: It is in the OZ contracts repo. But figured we should create the registry here.

chadoh (approved):
✨

PR #484, opened 2026-04-22
feat: deploy from external subregistries

willemneal:
## Summary
- Adds `deploy_with_subregistry` to the registry contract: deploys a wasm whose hash/version is resolved from a separate registry contract via a new `xcc_hash_and_version` cross-contract call.
- Extends the `Deploy` event with the source registry address and collapses subregistry xcc failures into a new `SubRegistryCrossContractCallFailed` error, while letting contract-returned errors (e.g. `NoSuchWasmPublished`, `NoSuchVersion`) propagate through.
- Wires the CLI's `deploy` command to route through `deploy_with_subregistry` when the wasm source registry differs from the local registry.
- **Pins the root registry address inside each subregistry at construction time** and changes `deploy_with_subregistry`'s `subregistry` argument from `Address` to `String` (a name). The contract resolves the name through the trusted root — closing the attack vector where a caller could pass any address and claim it was a known subregistry. The constructor's `is_root: bool` is replaced by `root: Option<Address>` to carry the pinned pointer.

## Test plan
- [x] `cargo test -p registry --lib` — unit tests for the happy path, missing wasm, missing version, non-registry target, and two independent subregistries in one env
- [x] CLI integration tests for the cross-registry deploy flow (all four pass: same-registry, root→sub, sub→root, sub→sub)
- [ ] Manual smoke test against a testnet with two registry contracts

### DIFF
--- a/contracts/registry/Cargo.toml
+++ b/contracts/registry/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "registry"
 description = "A crate for managing and deploying smart contracts on the Soroban blockchain."
-version = "0.5.1"
+version = "0.5.2"
 authors = ["The Aha Company <hello@theaha.co>"]
 license = "Apache-2.0"
 rust-version = "1.69"
--- a/contracts/registry/Cargo.toml
+++ b/contracts/registry/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "registry"
 description = "A crate for managing and deploying smart contracts on the Soroban blockchain."
-version = "0.5.2"
+version = "0.6.0"
 authors = ["The Aha Company <hello@theaha.co>"]
 license = "Apache-2.0"
 rust-version = "1.69"
--- a/contracts/registry/src/registry/wasm.rs
+++ b/contracts/registry/src/registry/wasm.rs
@@ -153,13 +153,11 @@ pub trait Publishable {
         wasm_name: soroban_sdk::String,
         version: Option<soroban_sdk::String>,
     ) -> Result<(soroban_sdk::String, soroban_sdk::BytesN<32>), Error> {
-        let version = version.map_or_else(
-            || Contract::most_recent_version(env, &wasm_name.clone().try_into()?),
-            Ok,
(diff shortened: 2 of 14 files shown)


contracts/registry/src/registry/wasm.rs
```diff
@@ -146,6 +146,23 @@ pub trait Publishable {
         Contract::get_hash(env, &wasm_name.try_into()?, version)
     }
 
+    /// Fetch the hash of a Wasm binary from the registry and bump TTL
```
chadoh:
```suggestion
    /// Fetch the hash and version of a Wasm binary from the registry and bump TTL
```

contracts/registry/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "registry"
 description = "A crate for managing and deploying smart contracts on the Soroban blockchain."
-version = "0.5.1"
+version = "0.5.2"
```
chadoh:
```suggestion
version = "0.6.0"
```

contracts/registry/src/registry/wasm.rs
```diff
@@ -153,13 +153,11 @@ pub trait Publishable {
         wasm_name: soroban_sdk::String,
         version: Option<soroban_sdk::String>,
     ) -> Result<(soroban_sdk::String, soroban_sdk::BytesN<32>), Error> {
-        let version = version.map_or_else(
-            || Contract::most_recent_version(env, &wasm_name.clone().try_into()?),
-            Ok,
-        )?;
+        let wasm_name: NormalizedName = wasm_name.try_into()?;
+        let version = Contract::get_version(env, &wasm_name, version)?;
```
chadoh, line 157: 💋

chadoh (approved):
Looks great! Would be nice to add a test showing what happens when a name is given that does not match anything in the root registry's lookup. (Right now we have tests for what happens when we give it a valid contract name when that contract is not a registry, but not what happens when we give it an invalid contract name.) But we don't _need_ that to merge this!

PR #491, opened 2026-04-29
Feat/pkg manager improvements

zachfedor:
Adds a bunch of improvements to the package manager feature:
- Fixes comma formatting issue when adding `packageManager` field to package.json
- Adds support for Bun
- Adds support for Deno
- Adds unit tests for new utilities supporting package manager choice (e.g. which are installed? what's listed in package.json?)
- Stifles a hardcoded print line including "npm" from stellar cli's contract binding command
- Adds a flag to specify package manager on init and skip prompt
- Adds a prompt to install missing extensions listed in environments.toml
- Adds a flag to skip all prompts on init and use defaults

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/init.rs
+++ b/crates/stellar-scaffold-cli/src/commands/init.rs
@@ -165,20 +165,17 @@ impl Cmd {
         let pm_command = pkg_manager.kind.command();
         let install_succeeded = run_install(pm_command, &absolute_project_path, &printer);
 
-        // Build contracts and create contract clients
-        printer.infoln("Building contracts and generating client code...");
-        // Use clap to parse build command with defaults, then configure programmatically
-        let mut build_command = build::Command::parse_from(["build", "--build-clients"]);
+        // Compile contracts only — client generation requires a live network and is handled by `start`
+        printer.infoln("Compiling contracts...");
+        let mut build_command = build::Command::parse_from(["build"]);
         build_command.build.manifest_path = Some(absolute_project_path.join("Cargo.toml"));
-        build_command.build_clients_args.env = Some(build::clients::ScaffoldEnv::Development);
-        build_command.build_clients_args.workspace_root = Some(absolute_project_path.clone());
         let mut build_args = global_args.clone();
         if !(global_args.verbose && global_args.very_verbose) {
             build_args.quiet = true;
         }
 
         if let Err(e) = build_command.run(&build_args).await {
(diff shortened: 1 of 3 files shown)


crates/stellar-scaffold-cli/src/commands/init.rs
```diff
@@ -123,19 +139,38 @@ impl Cmd {
...
 
-        let Some(pkg_manager) = select_pkg_manager(&printer) else {
+        let Some(pkg_manager) = (match &self.package_manager {
+            Some(kind) => {
+                let version = pkg_manager_version(kind.command());
+                Some(PackageManagerSpec {
+                    kind: kind.clone(),
+                    version,
+                })
+            }
```
willemneal, line 149: This feels like it should be an `impl From<PackageManager> for PackageManagerSpec`. Or what makes more sense to me is that `version` is a method on package manager. Below you keep using `.kind`

crates/stellar-scaffold-cli/src/commands/init.rs
```diff
@@ -123,19 +139,38 @@ impl Cmd {
...
+            None => select_pkg_manager(&printer, self.yes),
+        }) else {
```
willemneal, line 151: If you did you could just have this.
```suggestion
        let Some(pkg_manager) = self.package_manager.as_ref()
                                                                                            .or_else(||select_pkg_manager(&printer, self.yes)) 
         else {
```

crates/stellar-scaffold-cli/src/commands/init.rs
```diff
@@ -357,6 +389,116 @@ fn run_install(pm_command: &str, path: &Path, printer: &Print) -> bool {
...
+        .map(|e| e.name)
+        .collect();
+
+    if missing.is_empty() {
+        return;
+    }
+
+    let (known, unknown): (Vec<_>, Vec<_>) = missing
+        .iter()
+        .partition(|name| KNOWN_EXTENSIONS.iter().any(|(k, _)| k == name));
```
willemneal, line 437: Itertools has a `.unique()` and is already a dep. Also here you keep collecting and then immediately converting back into an iter. Perhaps `list_extensions` can be generic an the T implements IntoIter, this would let you keep the the other places while also letting you pass an iter here.

willemneal (approved):
There are some places for nits for improvement, but I think that can happen in a further rewrite. Nothing blocking just style and some optimizations.

PR #509, opened 2026-05-12
feat: pilot quorum + on-chain submit workflows

willemneal:
Add the two workflows that close the loop from in-review issue to
on-chain registry transaction, plus the per-kind quorum policy and
the CLI-args parser:

- .github/registry-quorum.yml: per-kind `min_voters` and
  `require_unanimous` (admin kind requires unanimous; any 👎 blocks).
  Default formula mirrors SCF pg-atlas: `ups ≥ min_voters + 2*downs`.
- registry-quorum-check.yml: triggered on `.quorum-check` comments,
  allowlisted to the `registry-pilots` team via a read-org PAT.
  Tallies pilot 👍/👎 reactions against the policy, swaps
  in-review → accepted/rejected, dispatches the on-chain workflow on
  accept, closes+locks the issue on reject.
- registry-onchain-submit.yml: workflow_dispatch only. Re-fetches and
  re-validates the issue body (defense in depth), parses to CLI args
  via parse_intake_to_args.py, runs stellar-registry-cli from a
  protected GitHub Environment (`registry-testnet` / `registry-mainnet`)
  that gates the signer secret behind required reviewers, posts the tx
  hash back, labels `:submitted` or `:submission-failed`.
- parse_intake_to_args.py: single source of truth for kind → CLI
  subcommand + flags mapping; emits a download URL when the publish
  method is selected (workflow fetches the wasm before invoking).

Tied together, these expect the operator to have already created the
`registry-pilots` team, set `READ_ORG_MEMBERS_PAT`, generated a CI
keypair, called `set_manager(<CI_PUBKEY>)` once per network as registry
admin, and provisioned the `registry-{testnet,mainnet}` environments
with secrets and reviewers. See docs/ci-publishing.md (follow-up
commit) for the runbook.

Testable after merge (with bootstrap done): pilots vote on an
in-review issue, one comments `.quorum-check`, the on-chain workflow
pauses at the env gate, a reviewer approves, the tx hash is posted
back to the issue.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.github/ISSUE_TEMPLATE/registry-admin.yml
+++ b/.github/ISSUE_TEMPLATE/registry-admin.yml
@@ -1,0 +1,111 @@
+name: "Registry: Admin op (manage / upgrade contract)"
+description: Request a sensitive op — change owner/address/name of a registered contract, or upgrade a deployed contract.
+title: "Registry Admin: "
+labels: ["registry-intake", "registry-intake:admin"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        # Registry Admin Intake (high-sensitivity)
+
+        Use this form for sensitive operations on already-registered or already-deployed
+        contracts. **Higher quorum applies** (see `.github/registry-quorum.yml`).
+
+        Only fill in the fields that apply to your chosen `method`. The validator will
+        reject submissions where required fields for that method are missing.
+
+        > **Out of scope for v1:** `set_admin`, `set_manager`, `remove_manager`, and
+        > registry-self `upgrade` all require **registry admin** auth, not the CI key.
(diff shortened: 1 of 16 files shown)


willemneal:
Closing in favor of #518 — registry governance is moving on-chain via the Tansu-DAO-gated registry manager rather than the GitHub IssueOps quorum approach.

Issue #419, opened 2026-03-18
`import_contract!` macro


chadoh:
For any smart contract in Stellar Registry, we want to make cross contract calls as easy, approximately, as:

```rs
pub fn thing_doer(env: Env) {
    let dao = stellar_registry::import_contract!(env, our_dao);
    dao.create_proposal(
        ...
    );
}
```

Where `our_dao` is a possible name of a contract deployed on Stellar Registry. Let's suppose it's [a Tansu DAO](https://github.com/Consulting-Manao/tansu/blob/21d68ee1d8f9567a1c9047f9eae2b42432ec8f22/contracts/tansu/src/contract_dao.rs#L171-L181). 

# Tough things

When you build your contract, you would **need to specify network**, testnet or mainnet. The macro would need to make network calls to look up the contract in the appropriate registry.

PR #518, opened 2026-05-19
feat: tansu-DAO-gated registry manager contract

willemneal:
## Summary

Adds `contracts/registry-tansu-manager/`, a Soroban contract that gates manager-authed operations on the on-chain registry behind approved Tansu DAO proposals. Single entry point: `trigger(proposal_id)`. Single transaction, end-to-end:

```text
user.tx
  └── manager.trigger(proposal_id)
        ├── read proposal from Tansu under our project_key       ← gate
        ├── env.authorize_as_current_contract(outcome_call)      ← pre-auth
        └── invoke_contract(Tansu, "execute", [self, project_key, id, _, _])
              ├── auth_maintainers: manager.require_auth         ← implicit (direct caller)
              ├── tally votes → flip status → Approved
              └── try_invoke_contract(outcome)                   ← registry.publish_hash/deploy(args)
                    └── manager.require_auth                     ← matched by pre-auth
                          → outcome runs in the same tx
```

Key properties:

- **Config (`tansu`, `project_key`, `registry`) is set immutably in `__constructor`.** Wrong-project callers can't piggyback — `trigger` reads the proposal under *our* configured `project_key` and pre-authorizes only that one outcome.
- **The manager must be the Tansu project's sole maintainer** (set up at deploy time via `Tansu::register(…, maintainers=[manager])` or `update_config`). That way the manager is the direct caller of `Tansu.execute`, so Tansu's internal `maintainer.require_auth()` is satisfied by Soroban's contract-implicit auth — no auth entry needed and no `enable_non_root_authorization` simulator dance.
- **The registry's `manager.require_auth()` is satisfied by [`env.authorize_as_current_contract`](https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Env.html#method.authorize_as_current_contract).** The pre-auth is scoped to one specific `(contract, fn, args)` triple — nothing else gets authorized.
- **No on-manager replay guard.** Tansu's own `if proposal.status != Active { panic }` (rejected on replay with `#402 ProposalActive`) is the source of truth.

Tansu's `Proposal` / `OutcomeContract` / `ProposalStatus` types are derived at compile time from `tansu-stub`'s wasm spec via `stellar_registry::import_contract_client!(tansu_stub)` — `tansu-stub` carries the canonical contracttype layout, mirroring upstream `Consulting-Manao/tansu`. We'd have liked to `contractimport!` the live Tansu wasm directly, but its v2.0.2 spec emits both a `pub trait Contract` (the interface) and a `#[contracttype] pub struct Contract` (used as a typed-Address wrapper in `set_nqg_contract` etc.) into the same generated module, which collides — tracked in [`Consulting-Manao/tansu#152`](https://github.com/Consulting-Manao/tansu/issues/152). The stub also gains a Tansu-shaped `execute(maintainer, project_key, proposal_id, _, _) -> ProposalStatus` that auto-invokes the outcome the way real Tansu does, so harnesses that target the stub exercise the full `trigger` chain.

Supporting pieces:
- `contracts/hello/` — minimal hello-world contract used as the payload by all three e2e scripts.
- `contracts/test/tansu-stub/` — stub Tansu (now with `execute` + a `#402 ProposalActive` error to mirror Tansu's replay guard).

## Test plan

- [x] `cargo test -p registry-tansu-manager` — 1 constructor smoke test. The integration behavior is covered by the e2e scripts below.
- [x] `cargo test -p hello -p tansu-stub` — passes.
- [x] `cargo clippy -p registry-tansu-manager -p tansu-stub --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.
- [x] `just build` — wasm builds for `registry`, `registry_tansu_manager`, `tansu_stub`, and `hello`. Topo sort orders `tansu-stub` before the manager via `[package.metadata.stellar] contract = true`.
- [x] **Stub-based testnet** via `contracts/registry-tansu-manager/e2e-testnet.sh` — see below.
- [x] **Custom-Tansu testnet** via `contracts/registry-tansu-manager/e2e-fast-tansu-testnet.sh` (~2.5min wall-clock vs ~48h on live Tansu) — see below.
- [ ] **Live-Tansu testnet** via `contracts/registry-tansu-manager/e2e-real-tansu-testnet.sh` — phase 1 has run; phase 2 against live Tansu hasn't been re-verified post-pivot (24h voting + 24h execute timelock).

## Stub-based testnet verification (`e2e-testnet.sh`)

Ran `NETWORK=testnet ./contracts/registry-tansu-manager/e2e-testnet.sh` end-to-end. All steps green:

1. Deploy fresh registry (admin as initial manager so the publisher can self-publish).
2. Upload `hello.wasm` and have admin-as-manager publish it on the author's behalf.
3. Deploy `tansu-stub`.
4. Deploy `registry-tansu-manager` pointing at stub + registry.
5. Admin swaps the registry's manager to the manager contract.
6. Plant an `Approved` deploy-proposal on the stub (outcome targets `registry.deploy(...)`).
7. `manager.trigger(proposal_id)` — pre-authorizes the outcome via `authorize_as_current_contract`, calls `stub.execute(self, project_key, proposal_id, _, _)`; the stub auto-invokes the outcome; `registry.deploy` runs under the pre-auth; hello deploys.
8. `hello(world) → "world"` on the freshly deployed contract.
9. Re-`trigger` → `#402 ProposalActive` from the stub (mirroring Tansu's status guard).

Contracts deployed during the verification run:

| Role | Address |
| ---- | ------- |
| registry | [`CA72QP7N6RTTMRYLMT7PUSYPWRBJGWMR5YOFDOSNTUOTX63SYNKU4445`](https://stellar.expert/explorer/testnet/contract/CA72QP7N6RTTMRYLMT7PUSYPWRBJGWMR5YOFDOSNTUOTX63SYNKU4445) |
| registry-tansu-manager | [`CAG7UKI6FCZB6464GUDE2FRVEOZR5QRIZACNVNSQSA7UXTFMN3ZQMKRI`](https://stellar.expert/explorer/testnet/contract/CAG7UKI6FCZB6464GUDE2FRVEOZR5QRIZACNVNSQSA7UXTFMN3ZQMKRI) |
| tansu-stub | [`CCBXY322ZYY4RLS3WEBD7BZPTMR4WEDGNTIS3OXTWVVJ2M4IG4OTRW2U`](https://stellar.expert/explorer/testnet/contract/CCBXY322ZYY4RLS3WEBD7BZPTMR4WEDGNTIS3OXTWVVJ2M4IG4OTRW2U) |
| hello (deployed via proposal) | [`CATBV7SZS6Q57SKHKQWAGT3D2CBQQ522NUTZFKJPGTZJUSPYNKEBIIF4`](https://stellar.expert/explorer/testnet/contract/CATBV7SZS6Q57SKHKQWAGT3D2CBQQ522NUTZFKJPGTZJUSPYNKEBIIF4) (registered as `hello-1780000169`) |

`manager.trigger` tx (single-tx flow including stub.execute → registry.deploy): [`4c0bff33…ca87d`](https://stellar.expert/explorer/testnet/tx/4c0bff331598077eea0f50154a6e1dcd157cc4977d135b5b0318367c772ca87d).

## Fast Tansu-fork testnet verification (`e2e-fast-tansu-testnet.sh`)

A custom build of Tansu (same `Proposal`/`OutcomeContract` wire shape; adds `min_voting_period: Option<u64>` and `execute_delay: Option<u64>` to `register`) is deployed at [`CDK7JBIIP6E75HOYLGRGWAHQLT6JUNUXQ7GNOYS3NAP26GISUXJ26UON`](https://stellar.expert/explorer/testnet/contract/CDK7JBIIP6E75HOYLGRGWAHQLT6JUNUXQ7GNOYS3NAP26GISUXJ26UON) on testnet. This collapses the ~48h voting+timelock window into the script-configurable `MIN_VOTING_PERIOD` + `EXECUTE_DELAY` (defaults: 60s + 60s).

End-to-end against that Tansu:

1–4. Register fresh project (`min_voting_period=60`, `execute_delay=60`), add members, upload hello.wasm, deploy registry, deploy manager.
5. `set_manager(manager)` on the registry, `update_config(maintainers=[manager])` on Tansu so the manager is the sole project maintainer.
6. Create proposal whose outcome is `registry.publish_hash(hello, …)`.
7. Vote Approve, wait through voting + execute_delay (~2 min).
8. `manager.trigger(proposal_id)` — one tx flips the proposal Approved, runs the outcome, registry publishes. Verify with `registry.fetch_hash`.
9. Replay → `#402 ProposalActive`.

`manager.trigger` tx end-to-end (Tansu.execute + Proposal collateral refund + outcome auto-invocation + registry.publish_hash, all in one): [`df2c8bab…ea360`](https://stellar.expert/explorer/testnet/tx/df2c8baba55069516757e7222eddc0b493acd49cd6a2d7ead5dbb7ed855ea360).

## Live-Tansu testnet verification (`e2e-real-tansu-testnet.sh`)

Two-phase script that exercises the same flow against the **live** Tansu DAO on testnet ([`CBXKUSLQPVF35FYURR5C42BPYA5UOVDXX2ELKIM2CAJMCI6HXG2BHGZA`](https://stellar.expert/explorer/testnet/contract/CBXKUSLQPVF35FYURR5C42BPYA5UOVDXX2ELKIM2CAJMCI6HXG2BHGZA)) — real `register` / `add_member` / `create_proposal` / `vote`, full 24h voting + 24h timelock window:

- `./e2e-real-tansu-testnet.sh setup` — registers project, auto-registers `.xlm` domain on SorobanDomain, deploys registry + manager, `update_config` to hand maintainership to the manager, `create_proposal` with `registry.publish_hash(...)` outcome, vote Approve, save state.
- `./e2e-real-tansu-testnet.sh finalize <state-file>` — after voting + timelock ends, calls `manager.trigger(proposal_id)` and asserts `registry.fetch_hash`.

Phase 1 was verified on 2026-05-28 (see earlier commit messages). The trigger pivot landed today; phase 2 against live Tansu hasn't been re-run end-to-end yet (requires a fresh setup + 48h wait).

## Out of scope (deliberate)

- **Multi-outcome proposals.** Rejected with `MultipleOutcomes` for v1.
- **`try_invoke_contract` on the registry call from inside the manager.** With `authorize_as_current_contract`, the registry's call is invoked by Tansu (not the manager), and a failure there propagates through Tansu's own `try_invoke_contract` as `#403 OutcomeError`. Acceptable for v1.
- **Pure custom-account `__check_auth` design.** Equivalent security model but blocked by stellar-cli 26.0.0 not exposing `enable_non_root_authorization` (recording-mode simulation can't synthesize a non-root auth entry for a custom account whose `__check_auth` would satisfy it). Revisit when CLI support lands; the `trigger` design works with current tooling and standard frontend SDKs.


contracts/registry-tansu-manager/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+            .ok_or(Error::NoOutcomeContracts)?;
+        if outcomes.len() != 1 {
+            return Err(Error::MultipleOutcomes);
+        }
+        let oc = outcomes.get(0).unwrap();
+        if oc.address != registry {
+            return Err(Error::OutcomeTargetMismatch);
+        }
+
+        let result: Val = env.invoke_contract(&registry, &oc.execute_fn, oc.args);
```
tupui: Do we want to have any checks on the function which is allowed to be called?

willemneal: My thoughts here was that the proposal is for this contract to call any method on the registry. Then it's up to the contract to authorize it.

Honestly we could even drop the requirement for only calling the registry, but wanted to be more specific this time.

tupui: ok ok 👍

chadoh:
> Also lands two supporting pieces for the e2e test:
>
> - `contracts/hello/` — minimal hello-world contract used as the published+deployed payload.

Seems like this should live in `contracts/test` or some other test-specific directory

contracts/registry-tansu-manager/e2e-testnet.sh
```diff
@@ -0,0 +1,195 @@
...
+
+for w in "$HELLO_WASM" "$REGISTRY_WASM" "$MANAGER_WASM" "$STUB_WASM"; do
+    if [ ! -f "$w" ]; then
+        echo "❌ missing $w — run \`just build\` first" >&2
+        exit 1
+    fi
+done
+
+# Ensure the network alias exists locally. testnet is added if missing; any
+# other name must be pre-configured by the user.
```
chadoh: This isn't true, there's a bunch of default aliases since forever (a couple years at least?); bot doesn't know what it's talking about

```suggestion
# Ensure the network alias exists locally. 
```

contracts/registry-tansu-manager/e2e-testnet.sh
```diff
@@ -0,0 +1,195 @@
...
+#   4. Deploy the registry-tansu-manager, pointing at the stub + registry.
+#   5. Admin installs the manager on the registry.
+#   6. Plant an `Approved` deploy-proposal on the stub.
+#   7. Call manager.execute(proposal_id) — registry deploys hello via XCC.
+#   8. Verify: invoke hello on the freshly deployed contract.
+#   9. Replay guard: second execute(proposal_id) returns AlreadyExecuted.
+#
+# Usage: contracts/registry-tansu-manager/e2e-testnet.sh
+# Env vars:
+#   NETWORK         Stellar network alias (default: testnet; must be `stellar network add`-ed).
```
chadoh:
```suggestion
#   NETWORK         Stellar network alias (default: testnet; must be in `stellar network ls`).
```

contracts/registry-tansu-manager/e2e-testnet.sh
```diff
@@ -0,0 +1,195 @@
...
+# other name must be pre-configured by the user.
+if ! stellar network ls 2>/dev/null | grep -qx "$NETWORK"; then
+    if [ "$NETWORK" = "testnet" ]; then
+        stellar network add testnet \
+            --rpc-url https://soroban-testnet.stellar.org \
+            --network-passphrase "Test SDF Network ; September 2015"
+    else
+        echo "❌ stellar network '$NETWORK' is not configured; run \`stellar network add\` first" >&2
+        exit 1
+    fi
```
chadoh: Don't need to add it, it's there by default

```suggestion
    echo "❌ stellar network '$NETWORK' is not configured; run \`stellar network add\` first" >&2
    exit 1
```

contracts/registry-tansu-manager/e2e-testnet.sh
```diff
@@ -0,0 +1,195 @@
...
+echo "    hash:     $HELLO_HASH"
+
+echo "==> Publishing hello@$HELLO_VERSION (author=$AUTHOR_ADDR, manager=$ADMIN_ID)"
+stellar contract invoke --id "$REGISTRY_ID" \
+    --source "$ADMIN_ID" --network "$NETWORK" \
+    -- publish_hash \
+    --wasm_name hello \
+    --author "$AUTHOR_ADDR" \
+    --wasm_hash "$HELLO_HASH" \
+    --version "$HELLO_VERSION"
```
chadoh, line 111: Should these be combined to use `stellar registry publish` instead? Or to at least call `invoke -- publish` directly, instead of `invoke -- publish_hash`?

contracts/registry-tansu-manager/e2e-testnet.sh
```diff
@@ -0,0 +1,195 @@
...
+# 4. Manager pointing at the stub + registry.
+echo "==> Deploying registry-tansu-manager"
+MANAGER_ID=$(stellar contract deploy --wasm "$MANAGER_WASM" \
+    --source "$ADMIN_ID" --network "$NETWORK" \
+    --alias "manager-e2e-${RUN_ID}" \
+    -- \
+    --tansu "$TANSU_ID" \
+    --project_key "$PROJECT_KEY" \
+    --registry "$REGISTRY_ID")
+echo "    manager:  $MANAGER_ID"
```
chadoh, line 129: Also strikes me as a little odd that registry tests wouldn't use registry for all of these wasm uploads & contract deploys. Like, sure, we don't **need** to. But we've got a nifty naming tool, why aren't we using it?

That said, if it keeps the tests simpler/more minimal/focused to skip that, I'm fine keeping as-is.

contracts/test/tansu-stub/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+use soroban_sdk::{
+    contract, contractimpl, contracttype, vec, Address, Bytes, BytesN, Env, IntoVal, String,
+    Symbol, Val, Vec,
+};
+
+// Tansu Proposal types — kept in lock-step with both `Consulting-Manao/tansu`
+// `contracts/tansu/src/types.rs` and `contracts/registry-tansu-manager/src/lib.rs`.
+// Duplicated here (rather than path-dep'd) because linking the manager crate as
+// an `rlib` would re-export its `#[contractimpl]` functions into this stub's
+// wasm.
```
chadoh, line 13: Can you explain why "re-exporting its `#[contractimpl]` functions into this stub's wasm" is a bad thing? To me it seems possibly preferable to duplicating and "keeping in lock-step" (how are we actually planning to keep these in lock-step?)

contracts/test/tansu-stub/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tansu-stub"
+description = "Minimal stand-in for the Tansu DAO contract. Used by the e2e script for the registry-tansu-manager flow on testnet — it lets a caller plant a synthetic `Approved` proposal so `RegistryTansuManager::execute(proposal_id)` can be exercised without running Tansu's real voting cycle (collateral, members, 24h voting period)."
```
chadoh, line 3: I wonder if using the real tansu would be more robust. The 24hr voting period is the real dealbreaker, but perhaps we can update Tansu to make this configurable via environment variable. Then we can compile a custom Tansu and deploy it as part of the `e2e` script. 

3. Create a new project
4. Add minimum number of voters
5. Then we're good to go?

How complex is that? 

How complex is it, compared to maintaining a long-lived stub in a separate repo that is (magically?) kept "in lock-step" with the real Tansu?

Cargo.toml
```diff
@@ -4,6 +4,8 @@ default-members = ["crates/*"]
 members = [
     "crates/*",
     "contracts/registry",
+    "contracts/registry-tansu-manager",
+    "contracts/hello",
     "contracts/test/*",
     "crates/stellar-scaffold-test/fixtures/contracts/*",
```
chadoh, line 10: Seems weird to me that we can't list `contracts/*` here. It makes me wonder if there's a more natural place to put these new tests, and everything in `contracts/test/*`.

But also w/e ¯\\\_(ツ)_/¯

chadoh (changes requested):
Started this review from the `e2e` test stuff and worked my way down. Would like to do another pass at the stuff at the top. But there's enough feedback here already that I figured I'd submit now.

contracts/test/tansu-stub/Cargo.toml
willemneal: I went back and forth on it. The tricky part with using a real contract is 1) the constructor is not simple, 2) won't require the rewrite as you say, and 3) adds a new dep.

My thoughts are that we will work on the contract itself later and it is better to just handle the current API with the stub (with one real one on testnet which shows a success with the 24 hours). Then once updated we can move it as a first class contract in the registry as we are planning and then handle the rewrite then which should be straight forward. 

Since we are the maintainers of Tansu the changes won't spring up on us and we can use this as a vehicle for testing.

Cargo.toml
willemneal: The issue is that `contracts/*` will fail on the `contracts/tests/*` because that folder doesn't have a Cargo.toml. I do agree we need to move the hello to the test folder.

contracts/test/tansu-stub/src/lib.rs
willemneal: This was a quick and dirty way. I think I'll update to use contract import instead. Basically when "importing" another contract you don't want to use an `rlib` because of the way the spec generation macros work. Essentially they all get pulled in even if your contract doesn't export those types just uses them. The preferred way is via Wasm imports so that you get the real spec and not need to copy/paste. I'll fix it.

contracts/registry-tansu-manager/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Proposal's outcome targets an address other than the configured registry.
+    OutcomeTargetMismatch,
+    /// Proposal has already been executed by this manager.
+    AlreadyExecuted,
+}
+
+// Tansu types mirrored from Consulting-Manao/tansu `contracts/tansu/src/types.rs`.
+// Field order, variant order, and `#[contracttype]` annotations must stay
+// identical to the upstream — Soroban encodes structs/enums by position, so
+// silent drift would surface as decode failures at runtime.
```
chadoh: We definitely need to circle back to this! In Q3 (or just: once we've completed all our Q2 goals) we need to start using Tansu as a dependency, rather than copy-pasting. Can you make an issue so we don't forget?

contracts/registry-tansu-manager/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+        let tansu = Storage::get_tansu(env).unwrap();
+        let project_key = Storage::get_project_key(env).unwrap();
+        let registry = Storage::get_registry(env).unwrap();
+
+        let proposal: Proposal = env.invoke_contract(
+            &tansu,
+            &Symbol::new(env, "get_proposal"),
+            vec![env, project_key.into_val(env), proposal_id.into_val(env)],
+        );
```
chadoh: Why are we doing cross-contract calls this manual way? Instead of using our own [`import_contract_client!` macro](https://crates.io/crates/stellar-registry)? Are we awaiting the completion of [`import_contract!`](https://github.com/theahaco/scaffold-stellar/issues/419) to further simplify the interface?

contracts/registry-tansu-manager/src/test.rs
```diff
@@ -0,0 +1,374 @@
...
+
+fn setup() -> Setup {
+    let env = Env::default();
+    let project_key = Bytes::from_slice(&env, &[7u8; 16]);
+    let tansu = env.register(TansuStub, ());
+
+    // Pre-compute the manager address so the registry can be constructed with it
+    // as `manager`. (Registers a transient placeholder, then registers the real
+    // manager contract at a deterministic address derived from arg hash — simpler:
+    // register the manager first, then the registry with the manager address.)
```
chadoh: Do I understand correctly that we think there's a simpler way to do this, but we are choosing to not do it? Why?

contracts/registry-tansu-manager/src/test.rs
```diff
@@ -0,0 +1,374 @@
...
+    );
+
+    let err = s.manager_client.try_execute(&1).err().unwrap().unwrap();
+    assert_eq!(err, Error::OutcomeTargetMismatch);
+}
+
+// ---------------------------------------------------------------------------
+// Auth-flow guard: the registry's manager-only function must reject calls
+// that come from somewhere other than the manager contract.
+// ---------------------------------------------------------------------------
```
chadoh: Seems like a misplaced comment (it is not relevant to the `approved_proposal_cannot_be_replayed` test below)

chadoh (dismissed):
A couple small questions/notes, but nothing blocking!

contracts/registry-tansu-manager/src/lib.rs
willemneal: Switched to `stellar_registry::import_contract_client!(tansu_stub)` — the manager now gets a typed `Client::get_proposal` and the `Proposal`/`OutcomeContract`/`ProposalStatus` types from the stub's wasm spec. The forward call to the registry (`env.invoke_contract(&registry, &oc.execute_fn, oc.args)`) stays untyped on purpose: an approved proposal targets *any* registry method, and a typed client can't express an arbitrary-method forward.

We can't import the real `Consulting-Manao/tansu` wasm yet — its v2.0.2 spec contains both `#[contract] pub struct Contract` and `#[contracttype] pub struct Contract` (a typed-`Address` wrapper used in `set_nqg_contract` etc.). `contractimport!` then generates a `pub trait Contract` colliding with the contracttype struct in the same module. `experimental_spec_shaking_v2` doesn't help (the type is reachable through the trait). Opened **Consulting-Manao/tansu#152** proposing the upstream rename — once it lands in a release we re-pin and drop the stub-as-source-of-truth.

`import_contract!` (#419) is still the right end state on top of this.

contracts/test/tansu-stub/src/lib.rs
willemneal: Now using `stellar_registry::import_contract_client!(tansu_stub)` from the manager — the stub crate is the single source of truth for the Tansu types, and the manager derives `Proposal`/`OutcomeContract`/`ProposalStatus` + a typed `get_proposal` `Client` from the stub's wasm spec (no rlib involved, so no `#[contractimpl]` re-export). One hand-copy remains — in this stub crate — mirroring the upstream `Consulting-Manao/tansu` types.

We tried to drop that copy too by importing the real Tansu wasm, but it has a `Contract` name collision (`#[contract] pub struct Contract` + `#[contracttype] pub struct Contract`) that crashes `contractimport!`. Filed **Consulting-Manao/tansu#152** asking for the rename; once it ships, the stub stops needing its own copy and can `contractimport!` the real wasm directly.

willemneal:
in favor of: https://github.com/stellar-registry/contracts/pull/5

PR #549, opened 2026-06-24
feat: upgrade to stellar protocol 27

zachfedor:
- Upgrade entire stack to Protocol 27: stellar-cli v27, soroban-sdk 27.0.0-rc.1, stellar-xdr 27.0.0, stellar-rpc-client 27.0.0
- Fix CI: update stellar/stellar-cli action to v27.0.0 in both build and test jobs
- Remove admin-sep dependency for test fixtures; rewrite hello_v1/hello_v2 fixtures with minimal inline upgrade logic
- Fix soroban-cli v27 breaking API changes


CLAUDE.md
```diff
@@ -57,7 +57,6 @@ Clippy and warning flags are configured in `.cargo/config.toml` `rustflags` and
 
 ### Other directories
 
-- `npm/` — npm wrapper that installs the prebuilt CLI binary
```
chadoh, line 60: Why did this go away?

zachfedor: It wasn't being used, it still pointed at stuff from Loam. I don't think it ever worked.

crates/stellar-scaffold-cli/build.rs
```diff
@@ -21,3 +23,36 @@ fn main() {
...
+/// Derive the Stellar protocol version the local network should run from the
+/// `stellar-cli` pin in the workspace `Cargo.toml`, and expose it as the
+/// `LOCAL_PROTOCOL_VERSION` compile-time env var. stellar-cli's major version
+/// tracks the protocol, so this stays correct across upgrades without a hardcoded
+/// number — bump the dependency and the local network follows.
+fn emit_local_protocol_version() {
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+    // crates/stellar-scaffold-cli -> workspace root is two levels up.
+    let workspace_toml = std::path::Path::new(&manifest_dir).join("../../Cargo.toml");
+    println!("cargo:rerun-if-changed={}", workspace_toml.display());
```
chadoh, line 36: Paging @willemneal on if this is the right way to do it. I thiiiink it's fine but want to make sure.

crates/stellar-scaffold-cli/build.rs
```diff
@@ -21,3 +23,36 @@ fn main() {
...
+
+    let version = parsed["workspace"]["dependencies"]["stellar-cli"]["version"]
+        .as_str()
+        .expect("stellar-cli workspace dependency must pin a version string");
+    // Strip any semver-requirement prefix (e.g. "=27.0.0") before taking the major.
+    let major = version
+        .trim_start_matches(['=', '^', '~', '>', '<', ' '])
+        .split('.')
+        .next()
+        .expect("stellar-cli version is empty");
```
chadoh: There might also be a way to do this that utilizes Cargo's internal version matching, rather than re-coding it via regex parsing, which seems brittle. @willemneal do you know?

chadoh (dismissed):
Nice work here, @zachfedor! This is way out of your typical wheelhouse and is some error-prone and frustrating work, but it looks like you found a way to make it less error-prone! 👏🏼 👏🏼 👏🏼 

I'd love a couple quick confirmations from @willemneal on my questions below before we merge.

crates/stellar-scaffold-cli/build.rs
willemneal: I'm less familiar with build.rs files but looks correct to me!

crates/stellar-scaffold-cli/build.rs
willemneal: Yep! https://crates.io/crates/semver

It will parse it and you can then access the major version. It will handle release candidates too if we ever decide to add a action that will test them out automatically.

PR #559, opened 2026-07-07
fix: read stellar-cli pin from manifest during cargo package

zachfedor:
Current release is held up by broken paths during cargo packaging

### DIFF
--- a/crates/stellar-scaffold-cli/build.rs
+++ b/crates/stellar-scaffold-cli/build.rs
@@ -25,28 +25,33 @@ fn main() {
 }
 
 /// Derive the Stellar protocol version the local network should run from the
-/// `stellar-cli` pin in the workspace `Cargo.toml`, and expose it as the
-/// `LOCAL_PROTOCOL_VERSION` compile-time env var. stellar-cli's major version
-/// tracks the protocol, so this stays correct across upgrades without a hardcoded
-/// number — bump the dependency and the local network follows.
+/// `stellar-cli` pin, and expose it as the `LOCAL_PROTOCOL_VERSION` compile-time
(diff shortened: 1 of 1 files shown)


crates/stellar-scaffold-cli/build.rs
```diff
@@ -57,3 +62,24 @@ fn emit_local_protocol_version() {
         .major;
     println!("cargo:rustc-env=LOCAL_PROTOCOL_VERSION={major}");
 }
+
+/// Read the `stellar-cli` version pin from a manifest, checking both the
+/// `[workspace.dependencies]` table (workspace root) and the `[dependencies]`
+/// table (a leaf crate, or a packaged crate with the pin inlined by Cargo).
+/// Returns `None` if the file is absent or holds no such pin.
+fn read_stellar_cli_version(path: &std::path::Path) -> Option<String> {
+    println!("cargo:rerun-if-changed={}", path.display());
```
chadoh, line 71: "cargo:rerun-if-changed"? What does that mean? Why are we printing it?

chadoh (approved):
Good find, and this looks like it will work fine as far as I understand it. I'd like @willemneal to chime in and let us know if there's a slicker, more Rusty way to do this.

crates/stellar-scaffold-cli/build.rs
zachfedor: It's a directive for Cargo, printing them is how you instruct it to build the package. It's super weird. I found them here: https://doc.rust-lang.org/cargo/reference/build-scripts.html

willemneal: I'm not very familiar with build.rs but i think this is a way to determine whether to rebuild the crate.

willemneal (approved):

PR #564, opened 2026-07-15
feat: allow --no-template option on init

### LINKED ISSUE #366: Allow users to choose "No Frontend" when scaffolding a new project
- [ ] create a new project without **any** frontend code
- [ ] prevent `scaffold build` from generating packages/clients

zachfedor:
Removes all JS-related files from repo during project instantiation, including any config files used by npm packages since they won't be installed. Also skips package manager choice since unneeded. Then sets `client = false` in environments.toml for all contracts.

Closes #366

### DIFF
--- a/crates/stellar-scaffold-cli/src/commands/init/instantiate.rs
+++ b/crates/stellar-scaffold-cli/src/commands/init/instantiate.rs
@@ -15,6 +15,24 @@ use crate::commands::{PackageManager, PackageManagerSpec};
 pub const TEMPLATES_DIR: &str = "templates";
 /// The directory selected template is promoted to
 pub const APP_DIR: &str = "app";
+/// Reserved `--template` value selecting the no-frontend layout. Not a
+/// directory under `templates/` — there is nothing to instantiate.
+pub const NO_FRONTEND: &str = "none";
+
+/// List of every JS-related file in the monorepo root that will be removed
+/// by the no-template CLI option. Only Rust/Cargo workspace remains.
+const NO_FRONTEND_REMOVALS: &[&str] = &[
+    TEMPLATES_DIR,
+    "app-lib",
+    "e2e",
+    "tests",
+    "scripts",
+    ".husky",
+    "node_modules",
+    "package.json",
+    "package-lock.json",
+    ".prettierignore",
+];
 
 /// The declared workspaces for the instantiated project. `e2e/` is kept (its
 /// self-adapting config targets the single `app/` post-init); `templates/*` is
@@ -39,6 +57,8 @@ pub enum Error {
     Io(#[from] std::io::Error),
     #[error("malformed root package.json: {0}")]
     PackageJson(#[from] serde_json::Error),
+    #[error("malformed environments.toml: {0}")]
+    EnvToml(#[from] toml_edit::TomlError),
 }
 
 /// How `--template` (or the interactive prompt) selected a source.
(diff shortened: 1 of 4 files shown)


crates/stellar-scaffold-cli/src/commands/init/instantiate.rs
```diff
@@ -106,6 +130,50 @@ pub fn instantiate(root: &Path, framework: &str) -> Result<(), Error> {
     Ok(())
 }
 
+/// Assemble the no-frontend layout: strip every JS workspace and node config
+/// from the acquired monorepo, leaving only the Cargo/contracts side, and set
+/// `client = false` on every contract in `environments.toml` so `build`/`watch`
+/// never generate (or deploy for) client packages there is no app to consume.
```
chadoh, line 136: I'm wondering if this is what @elliotfriend had in mind.

Elliot: we are adding a `--template none` option to Stellar Scaffold, so you can quickly scaffold a contracts-only project. You originally requested this feature. (It was a long time ago, and I lost track of the conversation. Was it in Slack? Was it in GitHub? Was it in Discord?)

Why use scaffold for this situation? I can think of several benefits, vs hand-rolling:

1. Toolchain consistency: quickstart, SDK version, CLI version, etc all wired together correctly
2. Rust workspace setup
3. Easy to add new OZ & `soroban-examples` contracts
4. Smarter builds: `stellar scaffold build` resolves inter-contract dependencies and builds in the correct order
5. **Easy integrations:** auto-generate TS bindings (aka NPM modules) for each of your contracts. Even if you're not using them in _this_ project, generating them as-you-go enables you to easily ship them for your users or export them for use in other projects.

Do we care about Benefit 5? Do you care, @elliotfriend?

If we do, then the functionality I'm commenting on here is wrong. We should not set `client: false` for each contract in `environments.toml`. We should keep creating clients.

Also, if we're going to skip creating clients, instead of setting `client: false`, we could remove `environments.toml` altogether (and/or remove `--build-clients` from the `package.json` scripts).

docs/site/docs/cli.md
```diff
@@ -13,13 +13,17 @@ stellar scaffold init <project-path> [name]
...
+- `--template <name>`: Template selector. A bare framework name (e.g. `react`) picks an official template; a `user/repo` shorthand (optionally with `#branch`) uses a community template; `none` creates a contracts-only project with no frontend. Omit to choose interactively
+- `--no-template`: Create a contracts-only project with no frontend (alias for `--template none`)
 
 The init command creates:
 
 - A new Stellar smart contract project with best practices and configurations
 - A frontend application using the [scaffold-stellar-frontend](https://github.com/stellar-scaffold/ui) template
 - Configuration files for both contract and frontend development
 
+With `--no-template` (or `--template none`), the frontend layer is omitted entirely: no `app/`, no JS workspaces, no dependency install. Every contract in `environments.toml` is written with `client = false`, so `stellar scaffold build` and `watch` skip client package generation. Flip a contract back to `client = true` if you later need TypeScript clients.
```
chadoh, line 25: I see, setting `client = false` for all of them makes it easy to later set `client = true`, if someone wants that. Seems reasonable enough! What do you think, @elliotfriend?

chadoh (approved):
Good enough for now! We can get feedback and see how people want to use this.

crates/stellar-scaffold-cli/src/commands/init/instantiate.rs
zachfedor: Reminder: setting `client: false` in `environments.toml` is just providing the default base for a post-init project. The user can always change those values after the fact and create the bindings if they'd like.

PR #569, opened 2026-07-21
feat: optimize build

chadoh:
stellar-cli v27 includes `--optimize` by default.

We just need to include `additional-libs` in order to enable it.

This makes the build for `stellar-scaffold-cli` much bigger!

Alternatives
------------

Rather than embedding `stellar-cli`, we could shell out to it. This would require a more significant re-architecture. Is it worth it, before shipping this feature? (Ideally, we should do some more data-driven comparison of the current approach vs status quo to understand the increased build/install time of `stellar-scaffold-cli` to see if we
think it's worth going with the simpler-to-ship approach for now.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


willemneal (approved):
Makes sense to me and we can shell out later if there are issues and in general move away from reimplementing the stellar cli

PR #578, opened 2026-08-05
Chore/cross repo gating

zachfedor:
Opt-in to cross-repo GH actions via a PR label instead of doing it silently to avoid accidental branch name collisions.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches: [main]
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
     paths-ignore:
       - "website/**"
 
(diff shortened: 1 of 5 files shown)


.github/workflows/rust.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches: [main]
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
chadoh, line 11: What is the “synchronize“ action? When does that get fired?

chadoh (approved):
Looks great!

.github/workflows/rust.yml
zachfedor: It's one of the defaults, I think it's when new commits are added. This change is only adding the `labeled` and `unlabeled` types to the other three that were already firing.

PR #588, opened 2026-09-08
feat: update Registry guide

### LINKED ISSUE #437: update scaffold tutorial with latest Registry
best practices, how it works, etc

chadoh:
Fixes https://github.com/stellar-scaffold/cli/issues/437

Stellar Registry has received significant updates since the initial
guide was written. This updates the guide to include the latest info.


docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
 # Registry Guide
 
-The Stellar Registry is a system for publishing, deploying, and managing smart contracts on the Stellar network. This guide explains how to use the registry CLI tools to manage your contracts.
+The Stellar Registry is a system for publishing, deploying, and managing smart contracts—and their underlying Wasms—on the Stellar network. This guide explains how to use the Registry CLI and UI to manage your contracts.
+
+<div class="videoWrapper">
+  <iframe src="https://www.youtube-nocookie.com/embed/xAlWmJOdMSQ?si=n2yYDkKbyqTAhiNP" title="Stellar Registry Full Walk-Through" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
+</div>
```
zachfedor, line 7: I think this is treated as JSX not HTML, so we should double check that the attributes are correct in the build (e.g. `className` vs `class`).

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
+
+2. **Contracts:** Once a Wasm is on-chain, anyone can deploy any number of contracts that all use that same underlying Wasm. If you're familiar with Object Oriented Programming, think of the Contract as the _instance_ and the Wasm as the _class._ A Wasm defines behavior; a Contract holds data (including a reference to the Wasm).
+
+   For a concrete example, think of a Wasm like [`oz/ft-standard`](https://stellar.rgstry.xyz/wasms/oz/ft-standard)—Open Zeppelin's standard Fungible Token implementation. You could create three separate tokens from this one Wasm just by deploying three separate smart contracts and giving them different initialization parameters (maybe you are a scammer and so you call them "USDC", "EURC", and "BTC"—nothing stops you from doing this). All these _Contracts_ reference the same _Wasm_ and thus have the same behavior, the same interface, the same methods; they only differ in their _storage_; in their initialization parameters and the various interactions (deposits, withdrawals, transfers) that adjust their stored data.
```
zachfedor: Can we come up with a better example than "maybe you're a scammer"? Maybe something legitimate?

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
-1. **On-chain registry contracts** - A root "verified" registry and an "unverified" registry
-2. The `stellar-registry` CLI tool for interacting with the registries
+### Quick Links
 
-### Registry Types
+Around the core smart contract, Stellar Registry wraps other tooling:
 
-There are two types of registries:
+- [Stellar Registry UI](https://rgstry.xyz), `rgstry.xyz`, where you can browse and search all Wasms and Contracts currently registered. [Source](https://github.com/stellar-registry/ui)
+- [Stellar Registry CLI](https://crates.io/crates/stellar-registry-cli), the main interface used to _write_ to Stellar Registry today. This is the tool you use to _publish Wasms_ and _deploy Contracts_ to Stellar Registry, or to register already-uploaded Wasms and already-deployd Contracts. [Source](https://github.com/stellar-registry/cli)
```
zachfedor:
```suggestion
- [Stellar Registry CLI](https://crates.io/crates/stellar-registry-cli), the main interface used to _write_ to Stellar Registry today. This is the tool you use to _publish Wasms_ and _deploy Contracts_ to Stellar Registry, or to register already-uploaded Wasms and already-deployed Contracts. [Source](https://github.com/stellar-registry/cli)
```

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
 Names in the registry support namespace prefixes. The CLI resolves names using the root registry as the source of truth:
 
 - `my-contract` - Looks up in the verified (root) registry
 - `unverified/my-contract` - First fetches the `unverified` registry contract ID from the root registry, then looks up `my-contract` in that registry
+- `some-other-subregistry/my-contract` — Same as above. Looks up `some-other-subregistry` in the root registry, then looks up `my-contract` in that registry.
 
 ### Name Normalization
 
-All names are normalized before storage:
+All names are normalized by the [stellar-registry-names](https://crates.io/crates/stellar-registry-name) crate before storage:
```
zachfedor:
```suggestion
All names are normalized by the [stellar-registry-name](https://crates.io/crates/stellar-registry-name) crate before storage:
```

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
+
+- [`oz/`](https://stellar.rgstry.xyz/wasms?query=oz), the Open Zeppelin Wasms (managed [by the Stellar Registry team](https://stellar.rgstry.xyz/wasms?query=oz))
+- [`circle/`](https://stellar.rgstry.xyz/contracts?query=circle), containing the official [`usdc`](https://stellar.rgstry.xyz/contracts/circle/usdc) and [`eurc`](https://stellar.rgstry.xyz/contracts/circle/eurc) tokens
+- `unverified/`: check out [Testnet Wasms](https://testnet.rgstry.xyz/wasms?query=unverified) and [Testnet Contracts](https://testnet.rgstry.xyz/contracts?query=unverified) as well as [Mainnet Wasms](https://stellar.rgstry.xyz/wasms?query=unverified) and [Mainnet Contracts](https://stellar.rgstry.xyz/contracts?query=unverified) to get a sense of how noisy this subregistry can be!
+
+When you first publish or deploy to Stellar Registry, you will need to use the `unverified` subregistry.
+
+Never this:
+
+```
```
zachfedor:
```suggestion
```bash
```

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
+# ❌ DOES NOT WORK
+stellar registry publish --wasm-name my-wasm
+```
+
+Instead, this:
+
+```
```
zachfedor:
```suggestion
```bash
```

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
+
+Some popular subregistries, to help you get a sense of how this all works:
+
+- [`oz/`](https://stellar.rgstry.xyz/wasms?query=oz), the Open Zeppelin Wasms (managed [by the Stellar Registry team](https://stellar.rgstry.xyz/wasms?query=oz))
+- [`circle/`](https://stellar.rgstry.xyz/contracts?query=circle), containing the official [`usdc`](https://stellar.rgstry.xyz/contracts/circle/usdc) and [`eurc`](https://stellar.rgstry.xyz/contracts/circle/eurc) tokens
+- `unverified/`: check out [Testnet Wasms](https://testnet.rgstry.xyz/wasms?query=unverified) and [Testnet Contracts](https://testnet.rgstry.xyz/contracts?query=unverified) as well as [Mainnet Wasms](https://stellar.rgstry.xyz/wasms?query=unverified) and [Mainnet Contracts](https://stellar.rgstry.xyz/contracts?query=unverified) to get a sense of how noisy this subregistry can be!
+
+When you first publish or deploy to Stellar Registry, you will need to use the `unverified` subregistry.
+
+Never this:
```
zachfedor: I feel like we should switch the order here, start with the happy path not the broken path.

docs/site/docs/registry.md
```diff
@@ -91,11 +166,11 @@ Options:
 
 Note: Use `--` to separate CLI options from constructor arguments.
 
-**Note:** For the verified registry, the manager must approve deploying with a registered name. For the unverified registry, use the `unverified/` prefix.
+**Note:** For the root registry, the governance process ([see above](#quick-links)) must approve initial publishes. For the unverified registry, use the `unverified/` prefix.
```
zachfedor: Two "Notes" in a row feels redundant. Also, do we even need the second one? It's copied verbatim from line 143, why the duplication?

Edit: wait, it's copied two more times on 216 and 239. Let's just do this once and call it out as an Admonition to make it more noticeable.

docs/site/docs/registry.md
```diff
@@ -91,11 +166,11 @@ Options:
...
 
 ### Deploy Unnamed Contract
 
-Deploy a published contract without registering a name in the registry. This is useful when you want to deploy a contract but don't need name resolution:
+Deploy a published contract without registering a name in the registry. This is useful when you want to deploy a contract but don't need name resolution. You can do this with the "Deploy a contract using this Wasm" button on any Wasm page in https://rgstry.xyz, or to use the CLI:
```
zachfedor: Let's link directly to the Wasms page here: 

```suggestion
Deploy a published contract without registering a name in the registry. This is useful when you want to deploy a contract but don't need name resolution. You can do this with the "Deploy a contract using this Wasm" button on any Wasm page in https://stellar.rgstry.xyz/wasms, or to use the CLI:
```

docs/site/docs/registry.md
```diff
@@ -1,31 +1,99 @@
...
 
-1. **On-chain registry contracts** - A root "verified" registry and an "unverified" registry
-2. The `stellar-registry` CLI tool for interacting with the registries
+### Quick Links
 
-### Registry Types
+Around the core smart contract, Stellar Registry wraps other tooling:
 
-There are two types of registries:
+- [Stellar Registry UI](https://rgstry.xyz), `rgstry.xyz`, where you can browse and search all Wasms and Contracts currently registered. [Source](https://github.com/stellar-registry/ui)
```
zachfedor, line 37: Should we use the full domain of `stellar.rgstry.xyz` here to be a bit more future-proof?

zachfedor (changes requested):
I wonder if some of this can be simplified if we move the more philosophical content to a new post on the [Blog](https://stellarscaffold.org/blog) (that looks a little bare at the moment). We're conflating our design decisions and architecture discussion with what could be a more straightforward pragmatic guide to using Registry.

docs/site/docs/registry.md
chadoh: Thanks for the warning! I double-checked and it looks like it works both ways! Keeping as-is.

docs/site/docs/registry.md
chadoh: I *think* root `rgstry.xyz` should be a little more future-proof? I'm basing this on prev experience when we changed `rgstry.xyz` from redirecting to `testnet.` to redirecting to `stellar.`

docs/site/docs/registry.md
chadoh: I sorta-blindly updated all previous "manager must approve" notes with this one. This part of this page turns into something like Reference Docs. If someone's looking at this part of the page, I'm not expecting them to necessarily scroll around and take in the rest of the page.

As for "Two Notes in a row": point taken, updated

